PR #10, opened 2018-04-11
asdf generation helpers and VISP changes

Cadair:
This is some changes needed to make the VISP test data work, and a couple of helper functions to make it a little easier.

### DIFF
--- a/dkist/io/reference_collections.py
+++ b/dkist/io/reference_collections.py
@@ -43,7 +43,10 @@ class BaseFITSArrayContainer(metaclass=abc.ABCMeta):
 
         self._check_contents(reference_array)
 
+        # If the first dimension is one we are going to squash it.
         reference_shape = reference_array.flat[0].shape
+        if reference_shape[0] == 1:
+            reference_shape = reference_shape[1:]
 
         self.shape = tuple(list(reference_array.shape) + list(reference_shape))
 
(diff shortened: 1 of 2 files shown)


dkist/asdf_maker/helpers.py
```diff
@@ -0,0 +1,182 @@
...
+    Given a FITS compliant header with CTYPEx,y as HPLN, HPLT return a
+    `~astropy.modeling.CompositeModel` for the transform.
+
+    This function finds the HPLN and HPLT keys in the header and returns a
+    model in Lon, Lat order.
+    """
+    latind = None
+    lonind = None
+    for k, v in header.items():
+        if isinstance(v, str) and "HPLN" in v:
```
Cadair: this should check for `-TAN` or else it should be smart about choosing the projection model below.

dkist/asdf_maker/helpers.py
```diff
@@ -0,0 +1,182 @@
...
+            lonind = int(k[5:])
+        if isinstance(v, str) and "HPLT" in v:
+            latind = int(k[5:])
+
+    if latind is None or lonind is None:
+        raise ValueError("Could not extract HPLN and HPLT from the header.")
+
+    # TODO: Add proper unit support
```
Cadair: Do this

dkist/asdf_maker/helpers.py
Cadair: done

PR #33, opened 2019-01-31
An experiment in removing astropy_helpers

Cadair:


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools", "wheel", "numpy"]
+requires = ["setuptools>=30.3.0", "wheel", "numpy"]
```
astrofrog: Include setuptools_scm here too?

setup.cfg
```diff
@@ -1,3 +1,49 @@
...
+github_project = DKISTDC/dkist
+classifiers =
+  'Programming Language :: Python'
+  'Programming Language :: Python :: 3'
+  'Programming Language :: Python :: 3.6'
+  'Programming Language :: Python :: 3.7'
+
+[options]
+python_requires = >=3.6
+setup_requires = setuptools_scm
```
astrofrog: I'd just get rid of this and include it in the toml file

Cadair: That would enforce pip 19? or only 18?

Cadair: Oh also, this would break `sdist` so that would be no good.

PR #41, opened 2019-02-27
Top Level ASDF Schema

### LINKED ISSUE #35: Implement top level asdf schema
It is possible with asdf to define a [top level schema](https://github.com/spacetelescope/asdf/pull/442) to validate the whole file. We should use this in the usertools to check that the Dataset asdf is compatible with our code.

### LINKED ISSUE #39: Add a binary table representation of all FITS headers to the asdf file
The table is already constructed to generate the asdf in [asdf_maker](https://github.com/DKISTDC/dkist/blob/master/dkist/asdf_maker/generator.py#L31) just need to put it in the asdf file in a way that means it's not all in the yaml, as that would slow the read down massively.

Cadair:
This adds the basic support for a [top level asdf schema](https://asdf.readthedocs.io/en/latest/asdf/features.html#custom-schemas) and adds a table of all the FITS headers, the values of the headers are stored as binary arrays.

Todo:

- [x] Improve top level tree naming. i.e. dataset -> data gwcs -> wcs.
- [x] Update test data with naming changes.
- [x] Make headers a required key and add it to all the test data.
- [x] Spec the top level `meta` key based on dataset inventory.

Closes #39 
Closes #35


dkist/dataset/tests/conftest.py
```diff
@@ -102,3 +102,10 @@ def dataset_4d(identity_gwcs_4d):
     array = da.from_array(x, tuple(shape))
 
     return Dataset(array, wcs=identity_gwcs_4d)
+
+
+if __name__ == "__main__":
```
Cadair: Delete me

PR #74, opened 2020-02-12
Improve performance of Dask array construction

Cadair:
The load only happens once per asdf file read, but we should still try and make it as efficient as possible.

### DIFF
--- a/dkist/io/array_containers.py
+++ b/dkist/io/array_containers.py
@@ -254,7 +254,7 @@ class DaskFITSArrayContainer(BaseFITSArrayContainer):
         """
         The `~dask.array.Array` associated with this array of references.
         """
-        return stack_loader_array(self.loader_array).reshape(self.output_shape)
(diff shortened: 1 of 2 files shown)


dkist/io/array_containers.py
```diff
@@ -253,7 +254,7 @@ def array(self):
         """
         The `~dask.array.Array` associated with this array of references.
         """
-        return stack_loader_array(self.loader_array).reshape(self.output_shape)
+        return stack_loader_array(loader_to_dask(self.loader_array)).reshape(self.output_shape)
```
Cadair: @astrofrog I can replace this line with:

```python
        return da.block(loader_to_dask(self.loader_array).tolist()).reshape(self.output_shape)
```
but then the chunksize of the output array is wrong:
```
dask.array<reshape, shape=(4, 32, 128, 128, 128), dtype=float64, chunksize=(1, 1, 128, 1, 128), chunktype=numpy.ndarray>
```

Cadair: And also it takes almost twice as long...

PR #87, opened 2020-12-09
Fix asdf write, also error on test warnings

Cadair:


dkist/net/tests/test_client.py
```diff
@@ -12,7 +12,7 @@
 from dkist.net.tests import strategies as dst  # noqa
 
 
-@pytest.fixture
+@pytest.fixture(scope='session')
```
Cadair, line 15:
```suggestion
@pytest.fixture
```

PR #102, opened 2021-02-26
Dataset search params

A-Derks:


dkist/net/tests/strategies.py
```diff
@@ -28,7 +28,7 @@ def _supported_attr_types():
     attr_types = list(walker.applymm.registry)
     attr_types.remove(object)
     attr_types.remove(AttrAnd)
-    attr_types.remove(a.dkist.BoundingBox)  # TODO: Support BoundingBox
+    #attr_types.remove(a.dkist.BoundingBox)  # TODO: Support BoundingBox
```
Cadair, line 31:
```suggestion
    attr_types.remove(a.dkist.BoundingBox)
```

dkist/net/attrs.py
```diff
@@ -278,3 +279,60 @@ def __init__(self, bottom_left, *, top_right=None, width: u.deg = None,
...
+
+    Parameters
+    ----------
+    spectralmin : `u.Quantity`
+        The minimum value of the average spectral sampling to search between.
+
+    spectralmax : `u.Quantity`
+        The maximum value of the average spectral sampling to search between.
+    """
+    def __init__(self, spectralmin: u.nm, spectralmax: u.nm):
```
Cadair, line 296: I *think* that we want to put [`u.quantity_input`](https://docs.astropy.org/en/latest/api/astropy.units.quantity_input.html?highlight=quantity_input) on these methods, and in this case enable the spectral equivalency:
```suggestion
   u.quantity_input(equivalencies=u.spectral())
    def __init__(self, spectralmin: u.nm, spectralmax: u.nm):
```

This should validate the units against the type annotations, we should add tests for this obviously.

dkist/net/attrs.py
```diff
@@ -278,3 +279,60 @@ def __init__(self, bottom_left, *, top_right=None, width: u.deg = None,
...
+
+    Parameters
+    ----------
+    spatialmin :
+        The minimum value of the average spatial sampling to search between.
+
+    spatialmax :
+        The maximum value of the average spatial sampling to search between.
+    """
+    def __init__(self, spatialmin, spatialmax):
```
Cadair, line 315: I think these should be in arcsec/pixel?

dkist/net/attr_walker.py
```diff
@@ -143,6 +147,20 @@ def _(wlk, attr, params):
     return params.update({'embargoEndDateMin': attr.start.isot,
                           'embargoEndDateMax': attr.end.isot})
 
+@walker.add_applier(SpectralSampling)
+def _(wlk, attr, params):
+    return params.update({'averageDatasetSpectralSamplingMin': attr.min,
+                          'averageDatasetSpectralSamplingMax': attr.max})
```
Cadair, line 153: These too will be quantities you will need to cast them back to float with `to_value`.

dkist/net/attr_walker.py
```diff
@@ -143,6 +147,20 @@ def _(wlk, attr, params):
...
+
+@walker.add_applier(SpatialSampling)
+def _(wlk, attr, params):
+    return params.update({'averageDatasetSpatialSamplingMin': attr.min,
+                          'averageDatasetSpatialSamplingMax': attr.max})
```
Cadair, line 158: Same here needs `to_value`.

Cadair (commented):
Looks like a good start, the results handling in `client.py` also needs updating so we handle these columns in the results table correctly as well.

dkist/net/tests/strategies.py
A-Derks: Why do we want to remove this from the list?

dkist/net/attr_walker.py
```diff
@@ -74,6 +74,10 @@ def _(wlk, attr, params):
         return params.update({'hasAllStokes': True})
     if attr.value.lower() == "intensity":
         return params.update({'hasAllStokes': False})
+    if attr.value.lower() == "spectral_axis":
+        return params.update({'hasSpectralAxis': True})
+    if attr.value.lower() == "temporal_axis":
+        return params.update({'hasTemporalAxis': True})
```
Cadair, line 80: This needs some thought (probably on my part). For a start I am not sure if you can AND two physobs together, where as the union of these two keys makes sense. Also I need to check the VSO data model to see if there is a neater way of doing this.

dkist/net/attrs.py
```diff
@@ -278,3 +279,61 @@ def __init__(self, bottom_left, *, top_right=None, width: u.deg = None,
...
+
+    Parameters
+    ----------
+    spatialmin :
+        The minimum value of the average spatial sampling to search between.
+
+    spatialmax :
+        The maximum value of the average spatial sampling to search between.
+    """
+    def __init__(self, spatialmin: u.arcsec/u.pix, spatialmax: u.arcsec/u.pix):
```
Cadair, line 316: add `u.quantity_input` here as well, but no equivalency is needed this time:
```suggestion
    @u.quantity_input
    def __init__(self, spatialmin: u.arcsec/u.pix, spatialmax: u.arcsec/u.pix):
```

dkist/net/attrs.py
```diff
@@ -278,3 +279,61 @@ def __init__(self, bottom_left, *, top_right=None, width: u.deg = None,
...
+
+    Parameters
+    ----------
+    temporalmin : `u.Quantity`
+        The minimum value of the average temporal sampling to search between.
+
+    temporalmax : `u.Quantity`
+        The maximum value of the average temporal sampling to search between.
+    """
+    def __init__(self, temporalmin: u.s, temporalmax: u.s):
```
Cadair, line 335:
```suggestion
    @u.quantity_input
    def __init__(self, temporalmin: u.s, temporalmax: u.s):
```

dkist/net/attr_walker.py
```diff
@@ -143,6 +147,20 @@ def _(wlk, attr, params):
     return params.update({'embargoEndDateMin': attr.start.isot,
                           'embargoEndDateMax': attr.end.isot})
 
+@walker.add_applier(SpectralSampling)
+def _(wlk, attr, params):
+    return params.update({'averageDatasetSpectralSamplingMin': attr.min.to_value(equivalencies=float),
+                          'averageDatasetSpectralSamplingMax': attr.max.to_value(equivalencies=float)})
```
Cadair, line 153: You need to give the target unit here.
```suggestion
    return params.update({'averageDatasetSpectralSamplingMin': attr.min.to_value(u.nm, equivalencies=u.spectral()),
                          'averageDatasetSpectralSamplingMax': attr.max.to_value(u.nm, equivalencies=u.spectral())})

```

dkist/net/attr_walker.py
```diff
@@ -143,6 +147,20 @@ def _(wlk, attr, params):
...
+
+@walker.add_applier(SpatialSampling)
+def _(wlk, attr, params):
+    return params.update({'averageDatasetSpatialSamplingMin': attr.min.to_value(equivalencies=float),
+                          'averageDatasetSpatialSamplingMax': attr.max.to_value(equivalencies=float)})
```
Cadair, line 158:
```suggestion
    return params.update({'averageDatasetSpatialSamplingMin': attr.min.to_value(u.arcsec/u.pixel),
                          'averageDatasetSpatialSamplingMax': attr.max.to_value(u.arcsec/u.pixel)})
```

Cadair (commented):
The test fails look like a weird interaction between the unit annotations and hypothesis. I will have to dive into that later.

PR #118, opened 2021-06-10
Update handling of asdfObjectKey

Cadair:


dkist/conftest.py
```diff
@@ -170,7 +170,7 @@ def dataset(array, identity_gwcs):
     meta = {'bucket': 'data',
```
Cadair, line 170: @JimArnow Does this test data look correct to you?

PR #127, opened 2021-08-17
Documentation

Cadair:
All documentation everywhere.


docs/index.rst
```diff
@@ -1,14 +1,44 @@
...
-  generated/gallery/index
+  guide/index
+  reference
   whatsnew/index
+
+
+First Steps
+-----------
+
+* **From the start**: :ref:`overview` | :ref:`install`.
```
fraserwatson, line 19: Is it intended for "overview" to link to the asdf docs?

docs/guide/level1data.rst
```diff
@@ -0,0 +1,30 @@
+.. _level1data:
+
+Level One DKIST Data
+====================
+
+The level one data provided by the DKIST data center has been calibrated to remove any effects introduced by the telescope or instruments.
+The result of these calibration recipes is a level one "dataset" which are the smallest units of DKIST data whic are searchable from the data center.
```
fraserwatson, line 7:
```suggestion
The result of these calibration recipes is a level one "dataset" which are the smallest units of DKIST data which are searchable from the data center.
```

docs/guide/level1data.rst
```diff
@@ -0,0 +1,30 @@
...
+
+Many FITS, one ASDF
+-------------------
+
+Due to the potential size of each of these datasets, and to eliminate on demand processing at the data center a single level one dataset is distributed across many FITS files each individual FITS file represents what can be considered to be a "single calibrated exposure".
```
fraserwatson, line 12:
```suggestion
Due to the potential size of each of these datasets, and to eliminate on demand processing at the data center a single level one dataset is distributed across many FITS files. Each individual FITS file represents what can be considered to be a "single calibrated exposure".
```

docs/guide/loading.rst
```diff
@@ -0,0 +1,85 @@
...
+* The `~dkist.Dataset.wcs` property is a `gwcs.wcs.WCS` object which provides coordinate information for all the pixels in the array.
+* The `~dkist.Dataset.files` property is an instance of the `dkist.io.FileManager` class which allows you to access and modify information about how the FITS files are loaded by the array.
+* The `~dkist.Dataset.headers` property is an `astropy.table.QTable` instance which is a complete record of all the FITS headers for all the FITS files in the dataset. This provides access to all the available metadata about the dataset without having to retrieve or open any of the FITS files.
+
+The relationship between `~dkist.Dataset.data` and `~dkist.Dataset.files`
+#########################################################################
+
+When an ASDF file containing a dataset is loaded, a new array is generated from the list of FITS filenames provided in that ASDF file.
+These FITS filenames are ordered in the correct way so that an array of the correct dimensionality to represent the whole dataset as a higher dimensional array that is contained in any FITS file.
```
fraserwatson, line 33:
```suggestion
These FITS filenames are ordered in the correct way so that an array of the correct dimensionality is created. This array represents the whole dataset as a higher dimensional array than is contained in any single FITS file.
```

fraserwatson (commented):
I think the biggest thing missing at this point would be an end to end walkthrough. Given an idea for a search (single instrument, time bounds) here is the code to:

1. search for the data
2. get the asdf
3. load it into user tools
4. download some or all of the data
5. slice along a non single-fits-file axis

I know the docs contain all of this info but I think it'll be really valuable to have an end to end listing of the code that would do it. And it isn't more than a few lines of code if I'm reading all this correctly.

docs/guide/level1data.rst
```diff
@@ -0,0 +1,30 @@
...
+The FITS files provided at level one by the data center, will conform to the FITS 4 specification and contain all the information about their individual array as well as DKIST specific keys in the header which specify its position in the larger dataset.
+To increase the usability of these data, in addition to all the FITS files (which could number many thousands) a single metadata only ASDF file is generated per dataset.
+
+The ASDF file provides the following information about the dataset:
+
+* A table of all the FITS headers for all the FITS files contained in the dataset.
+* An ordered nested list of the filenames of all the FITS files, providing the information about how to reconstruct the full dataset array from all the component FITS arrays.
+* Information about the dtype, shape and HDU number of the arrays in the FITS file.
+* A `gWCS <https://gwcs.readthedocs.io/>`__ object providing the coordinate information for the reconstructed dataset array.
+
```
A-Derks, line 29: Quality report?

A-Derks (commented):
+1 to Fraser's comment. I would also say something about the quality doc somewhere. That will be a part fo the asdf correct?

docs/guide/level1data.rst
Cadair: The quality report isn't in the asdf yet, so I will add it here when we modify the asdf.

docs/index.rst
Cadair:
```suggestion
* **From the start**: :ref:`install`.
```

Cadair: I removed the overview page.

Cadair:
@fraserwatson I would like to write a lot more examples, but as I can't actually do a search against the DC at the moment, I would rather wait until I can actually execute the code and generate a pretty gallery like the sunpy one?

PR #148, opened 2021-12-22
Implement models required for representing time varying VBI transforms

Cadair:
This PR (the first of a few) adds the models needed to represent the time varying pointing.

Notes to reviewers:

* This approach (putting these classes in here) will mean the inventory generation relies on the user tools
* The real magic is in [`CoupledCompoundModel`](https://github.com/DKISTDC/dkist/pull/148/files#diff-a9a4b9003939abfbec6973db1537eeb66005df3e86476b9285ba49be286350b3R224) and [`VaryingCelestialTransform`](https://github.com/DKISTDC/dkist/pull/148/files#diff-a9a4b9003939abfbec6973db1537eeb66005df3e86476b9285ba49be286350b3R137)
* `VaryingCelestialTransform` only supports the pointing changing over one pixel axis (this will need changing to support VISP).


Cadair:
pre-commit.ci autofix

dkist/wcs/models.py
```diff
@@ -0,0 +1,329 @@
...
+        # Pass through arguments to the left model unchanged while computing the right output
+        mapping = list(range(n_left_only_inputs))
+        step1 = m.Mapping(mapping) & right_inverse
+
+        # Now pass through the right outputs unchanged while also feeding them into the left model
+
+        # This mapping duplicates the output of the right inverse to be fed
+        # into the left and also out unmodified at the end of the transform
+        inter_mapping = mapping + list(range(max(mapping) + 1, max(mapping) + 1 + right_inverse.n_outputs)) * 2
+        step2 = m.Mapping(inter_mapping) | (left_inverse & m.Mapping(list(range(right_inverse.n_outputs))))
```
nden, line 327: Is this `m.Mapping(list(range(right_inverse.n_outputs)))` equivalent to `m.Identity(right_inverse.n_outputs)`?
If so it may improve readability.

dkist/wcs/models.py
```diff
@@ -0,0 +1,329 @@
...
+      └─┬───┬────┘      │
+        │   │           │
+        ▼   ▼           ▼
+        x   y           z
+
+    """
+
+    _separable = False
+
+    def __init__(self, op, left, right, name=None, shared_inputs=1):
```
nden, line 294: Is there a need to pass the `op` argument since it's essentially constant? Or just document it but keep the check?

dkist/wcs/models.py
```diff
@@ -0,0 +1,329 @@
...
+    def __init__(self, *args, crval_table=None, pc_table=None, projection=m.Pix2Sky_TAN(), **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.pc_table = np.asanyarray(pc_table)
+        self.crval_table = np.asanyarray(crval_table)
+        self.table_shape = self._validate_table_shapes(self.pc_table, self.crval_table)
+
+        if not isinstance(projection, m.Pix2SkyProjection):
+            raise TypeError("The projection keyword should be a Pix2SkyProjection model class.")
+        self.projection = projection
```
nden, line 126: Just a caution comment here - if any projection is allowed then care should be taken to compute `lon_pole`. This is not trivial in the general case. There should be a function in gwcs "soon" which retrieves the value from wcslib.

nden (commented):
@Cadair Looks reasonable to me :-)

dkist/wcs/models.py
Cadair: I didn't want to change the signature from `CompoundModel`, so I added the check. Also it's a little more explicit written out with the op?

dkist/wcs/models.py
Cadair: I don't think we will be using it for any projections other than TAN, but worth keeping an eye on thanks.

Cadair:
That oldestdeps fail needs fixing.

dkist/wcs/models.py
```diff
@@ -0,0 +1,340 @@
...
+    _separable = False
+
+    def __init__(self, op, left, right, name=None, shared_inputs=1):
+        if op != "&":
+            raise ValueError(
+                f"The {self.__class__.__name__} class should only be used with the & operator."
+            )
+        super().__init__(op, left, right, name=name)
+        self.n_inputs = self.n_inputs - shared_inputs
+        self.inputs = self.inputs[:-shared_inputs]
```
eigenbrot, line 312: It's likely I don't understand how compound models work, but I was surprised to see the implicit assumption that the shared inputs are the last in the `self.inputs` list. Given the docstring above I expected them to be somewhere in the middle.

dkist/wcs/models.py
```diff
@@ -0,0 +1,340 @@
...
+def generate_celestial_transform(crpix: Union[Iterable[float], u.Quantity],
+                                 cdelt: Union[Iterable[float], u.Quantity],
+                                 pc: Union[ArrayLike, u.Quantity],
+                                 crval: Union[Iterable[float], u.Quantity],
+                                 lon_pole: Union[float, u.Quantity] = None,
+                                 projection: Model = m.Pix2Sky_TAN()) -> CompoundModel:
+    """
+    Create a simple celestial transform from FITS like parameters.
+
+    Supports unitful or unitless parameters, but if any parameters have units
```
eigenbrot, line 29: Based on comments in the test it seems like a unitless set of inputs will be interpreted as degrees. Probably worth writing that down up here.

dkist/wcs/models.py
Cadair: The way the shared inputs work is that it's *all* the inputs to the right model also go into the left one, hence it's all the inputs at the end of the list.

Cadair: heh, which now I write out isn't actually enforced anywhere, I shall have a thing :smile:

dkist/wcs/models.py
```diff
@@ -0,0 +1,340 @@
...
+        The map projection to use, defaults to ``TAN``.
+
+    Notes
+    -----
+
+    This function has not been tested with more complex projections. Ensure
+    that your lon_pole is correct for your projection.
+    """
+    spatial_unit = None
+    if hasattr(crval[0], "unit"):
```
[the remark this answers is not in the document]
Cadair: I guess it is, but only because I can't specify a type hint for "a length 2 Quantity", it has to be length 2.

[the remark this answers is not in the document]
Cadair: I have added this to the docstring.

dkist/wcs/models.py
```diff
@@ -0,0 +1,340 @@
...
+                                            pc=pc,
+                                            crval=crval,
+                                            lon_pole=self.lon_pole,
+                                            projection=self.projection)
+
+    def evaluate(self, x, y, z, crpix, cdelt, lon_pole):
+        pc, crval = self.get_pc_crval(z, self.pc_table, self.crval_table)
+
+        sct = generate_celestial_transform(crpix=crpix[0],
+                                           cdelt=cdelt[0],
```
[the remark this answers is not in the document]
Cadair: The `evaluate` method is called by the depths of astropy, and I honestly have no idea why the shape of the inputs is always `(1, N)`.  I shall add a comment explaining it, but I wouldn't really be happy documenting the inputs in detail here as I am not sure I would get it right.

dkist/wcs/models.py
Cadair: I have now changed it around so that it drops the leading inputs from the right model. Thanks for the spot!

PR #149, opened 2022-01-27
Don't recursively test gwcs' serialisation

Cadair:


Cadair:
pre-commit.ci autofix

PR #152, opened 2022-02-01
Migrate to asdf Converters

Cadair:
This refactors the asdf code and schemas to use the new `Converter` interface introduced in asdf 2.8. It also takes the opportunity to clean up some historical choices, but maintains the ability to read old files.

~This is currently blocked on https://github.com/asdf-format/asdf/pull/1065~

TODO:

- [ ] ~Add a tiled dataset test data script and generate one for previous versions of the tag.~
- [x] Fix oldestdeps


dkist/io/asdf/converters/dataset.py
```diff
@@ -1,19 +1,18 @@
...
 
-from ..types import DKISTType
 
-__all__ = ["DatasetType"]
+class DatasetConverter(Converter):
+    tags = [
+        "tag:dkist.nso.edu:dkist/dataset-0.1.0",
+        "tag:dkist.nso.edu:dkist/dataset-0.2.0",
+        "tag:dkist.nso.edu:dkist/dataset-0.3.0",
+        "asdf://dkist.nso.edu/dataset-1.0.0",
```
eslavich, line 9: Elsewhere we've been putting tags into a /tags/ namespace (like with /schemas/).  This works just as well, but if you're interested in uniformity...

dkist/io/asdf/converters/file_manager.py
```diff
@@ -0,0 +1,34 @@
...
+        "tag:dkist.nso.edu:dkist/array_container-0.2.0",
+        "asdf://dkist.nso.edu/array_container-1.0.0",
+    ]
+    types = ["dkist.io.file_manager.FileManager"]
+
+    def from_yaml_tree(self, node, tag, ctx):
+        from dkist.io.file_manager import FileManager
+        from dkist.io.loaders import AstropyFITSLoader
+
+        filepath = Path((ctx.url or ".").replace("file:", ""))
```
eslavich, line 17: Maybe raise an error here if the URL has some other scheme besides `file:` ?  Just so the user gets a more specific message and not a failure down the line from trying to open an http URL as a file path.

setup.cfg
```diff
@@ -101,8 +101,17 @@ filterwarnings =
     ignore:the imp module is deprecated in favour of importlib:DeprecationWarning:ipykernel.iostream
     # Ignore warnings about asdf versions
     ignore:File.*asdf.extension.BuiltinExtension.*:asdf.exceptions.AsdfWarning
+    # Ignore asdf complaining about files written with old extension
+    ignore:File.*was created with extension class .*DKISTExtension.* which is not currently installed: asdf.exceptions.AsdfWarning
```
eslavich, line 105: There's a feature of the new-style extension that helps with this:

https://github.com/asdf-format/asdf/blob/master/asdf/extension/_extension.py#L39

dkist/io/asdf/converters/file_manager.py
```diff
@@ -0,0 +1,34 @@
+from pathlib import Path
+
+from asdf.extension import Converter
+
+
+class FileManagerConverter(Converter):
+    tags = [
+        "tag:dkist.nso.edu:dkist/array_container-0.2.0",
+        "asdf://dkist.nso.edu/array_container-1.0.0",
```
eslavich, line 9: If you like, you can register a YAML shorthand for your tag prefix:

https://github.com/asdf-format/asdf/blob/master/asdf/extension/_extension.py#L101

For example, if you use `!dkist!` as shorthand for `asdf://dkist.nso.edu/`, the ASDF file will be rendered like this:

```
!dkist!array_container-1.0.0
```

The only benefit is to reduce clutter when somebody pops open the file in an editor.

dkist/io/asdf/resources/schemas/dataset-0.1.0.yaml
```diff
@@ -12,7 +11,7 @@ description:
 type: object
 properties:
   data:
-    $ref: "http://dkist.nso.edu/schemas/dkist/array_container-0.2.0"
+    $ref: "asdf://dkist.nso.edu/schemas/file_manager-1.0.0"
```
eslavich, line 14: This this is a tagged object, consider using the `tag:` validator here

dkist/io/asdf/resources/schemas/dataset-0.2.0.yaml
```diff
@@ -12,7 +11,7 @@ description:
 type: object
 properties:
   data:
-    $ref: "http://dkist.nso.edu/schemas/dkist/array_container-0.2.0"
+    $ref: "asdf://dkist.nso.edu/schemas/file_manager-1.0.0"
```
eslavich, line 14: Consider `tag:` validator

dkist/io/asdf/resources/schemas/dataset-0.3.0.yaml
```diff
@@ -12,7 +11,7 @@ description:
 type: object
 properties:
   data:
-    $ref: "http://dkist.nso.edu/schemas/dkist/array_container-0.2.0"
+    $ref: "asdf://dkist.nso.edu/schemas/file_manager-1.0.0"
```
eslavich, line 14: Consider `tag:` validator

dkist/io/asdf/resources/schemas/tiled_dataset-0.1.0.yaml
```diff
@@ -17,7 +16,7 @@ properties:
     items:
       type: array
       items:
-        $ref: "http://dkist.nso.edu/schemas/dkist/dataset-0.3.0"
+        $ref: "asdf://dkist.nso.edu/schemas/dataset-0.3.0"
```
eslavich, line 19: Consider `tag:` validator here

eslavich (commented):
A smattering of suggestions, but nothing dramatic.  Looks good!

dkist/io/asdf/resources/schemas/tiled_dataset-0.1.0.yaml
Cadair: This one causes the validation with the top level schema to fail with 
```
E           jsonschema.exceptions.ValidationError: mismatched tags, wanted 'tag:dkist.nso.edu:dkist/dataset-0.3.0', got 'asdf://dkist.nso.edu/tags/dataset-1.0.0'
```
which I can not understand.

dkist/io/asdf/converters/file_manager.py
Cadair: Ohh, that's cool but I am not sure I really want to do it. I can't see many people opening the file and if they do the extra verbosity is probably helpful.

dkist/io/asdf/converters/file_manager.py
Cadair: In *theory* I would want the `FileManager` class to handle loading over http, so it should be ok if this returns something other than a file URL.

dkist/io/asdf/resources/schemas/tiled_dataset-0.1.0.yaml
eslavich: I think that's because you're using the tiled_dataset-0.1.0 schema for both tags in the manifests:

```yaml
  - schema_uri: "asdf://dkist.nso.edu/schemas/tiled_dataset-0.1.0"
    tag_uri: "tag:dkist.nso.edu:dkist/tiled_dataset-0.1.0"
```
```yaml
  - schema_uri: "asdf://dkist.nso.edu/schemas/tiled_dataset-0.1.0"
    tag_uri: "asdf://dkist.nso.edu/tags/tiled_dataset-1.0.0"
```

Since that's the case, you could use a combiner here:

```yaml
      items:
        oneOf:
          - tag: tag:dkist.nso.edu:dkist/dataset-0.3.0
          - tag: asdf://dkist.nso.edu/tags/dataset-1.0.0
```

(or write separate schemas, but the I think the combiner is a nice solution)

dkist/io/asdf/converters/file_manager.py
Cadair: In the end due to a *super* annoying windows bug I ended up erroring!

dkist/io/asdf/resources/schemas/tiled_dataset-0.1.0.yaml
Cadair: ah, I do actually have new schemas anyway, so thanks for spotting that!

PR #161, opened 2022-03-03
Varying celestial models with 2D lookup tables

Cadair:
This is a variant of the 3D model useful for VISP.

Todo:

- [x] Create a model which takes a 2D lookup table but only one extra index. Where the y pixel dimension is also used as the first index to the lookup table.
- [x] Also need a variant of the 1D table model where it takes no extra inputs and just varies with the y input.

### DIFF
--- a/dkist/io/asdf/tests/test_models.py
+++ b/dkist/io/asdf/tests/test_models.py
@@ -7,8 +7,11 @@ from astropy.coordinates.matrix_utilities import rotation_matrix
 from astropy.modeling import CompoundModel
 
 from dkist.wcs.models import (CoupledCompoundModel, InverseVaryingCelestialTransform,
-                              InverseVaryingCelestialTransform2D, VaryingCelestialTransform,
-                              VaryingCelestialTransform2D)
+                              InverseVaryingCelestialTransform2D,
+                              InverseVaryingCelestialTransformSlit,
+                              InverseVaryingCelestialTransformSlit2D, VaryingCelestialTransform,
+                              VaryingCelestialTransform2D, VaryingCelestialTransformSlit,
+                              VaryingCelestialTransformSlit2D)
 
 
 def test_roundtrip_vct():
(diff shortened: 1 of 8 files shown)


dkist/wcs/models.py
```diff
@@ -267,6 +276,279 @@ def evaluate(self, lon, lat, z, crpix, cdelt, lon_pole, **kwargs):
...
+        super().__init__(*args, **kwargs)
+        self.inputs = ("lon", "lat", "raster")
+        self.outputs = ("slit-y",)
+
+    def evaluate(self, lon, lat, y, crpix, cdelt, lon_pole, **kwargs):
+        return self._map_transform(lon, lat, y, crpix, cdelt, lon_pole,
+                                   inverse=True)[0]
+
+
+class VaryingCelestialTransformSlit2D(BaseVaryingCelestialTransform):
```
eigenbrot, line 461: I'm confused by the name of this model. It appears to be the same as `VaryingCelestialTransformSlit`, which already has 2 spatial dimensions, but with a time dimension added (the "repeat" input). Is that what the "Varying" implies?

dkist/wcs/models.py
```diff
@@ -394,14 +676,60 @@ def _calculate_separability_matrix(self):
...
         matrix[:, :left_solo_end] = cleft[:, :left_solo_end]
         matrix[:, left_solo_end:right_solo_start] = np.logical_or(cleft[:, left_solo_end:],
                                                                   cright[:, :self.shared_inputs])
-        matrix[:, right_solo_start:] = cright[:, :right_solo_start]
+        matrix[:, right_solo_start:] = cright[:, self.shared_inputs:]
 
         return matrix
+
+
+def varying_celestial_transform_from_tables(crpix: Union[Iterable[float], u.Quantity],
```
eigenbrot, line 692: Why `Iterable[float]`, but not `Iterable[u.Quantity]`?

dkist/wcs/models.py
```diff
@@ -267,6 +276,279 @@ def evaluate(self, lon, lat, z, crpix, cdelt, lon_pole, **kwargs):
...
+
+        # Put the units back
+        if isinstance(xx, u.Quantity):
+            x_out = x_out << xx.unit
+            y_out = y_out << yy.unit
+
+        return x_out, y_out
+
+
+class VaryingCelestialTransformSlit(BaseVaryingCelestialTransformSlit):
```
eigenbrot, line 413: After thinking about the comment on line 461 below this model name confuses me. Looking at the inputs it seems like "slity" is a detector dimension and "raster" is an index (see line 375 above). That sounds like 2 spatial dimensions to me and I don't see a time dimension despite "Varying" being in the name.

eigenbrot (approved):

dkist/io/asdf/tests/test_models.py
```diff
@@ -37,6 +41,63 @@ def test_roundtrip_vct():
...
+    assert u.allclose(world, (359.99861111, 0.00138889)*u.deg)
+
+    assert u.allclose(new_ivct(*world, 4*u.pix, 2*u.pix), pixel[:2], atol=0.01*u.pix)
+
+
+def test_roundtrip_vct_slit():
+    varying_matrix_lt = [rotation_matrix(a)[:2, :2]
+                         for a in np.linspace(0, 90, 10)] * u.arcsec
+
+    vct = VaryingCelestialTransformSlit(crpix=(5, 5) * u.pix,
```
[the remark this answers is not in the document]
Cadair: The 2D refers to the shape of the lookup table.

The non-slit models take one or two extra pixel inputs which are used to index the lookup table.

The slit variants of the models use one less extra pixel input, as the pointing varies along the second pixel dimension (step in the raster).

There is no mention of time here because all of these models only work in pixel space. Even for the inverse (world to pixel) transform the time coordinates are transformed to pixel and then passed to the inverse.

Hopefully that makes sense!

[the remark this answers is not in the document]
Cadair: Oh and also, these models only output spatial axes, the wave and temporal parts are done by other models.

dkist/wcs/models.py
Cadair: Because Quantity is a numpy array so already iterable.

PR #178, opened 2022-06-21
Fix path interpolation between download and fetch

### LINKED ISSUE #171: Unify path handling between `Fido` and Globus transfers
The path formatting in `Fido` and `FileManager.download` and `transfer_complete_datasets` isn't the same, which makes it hard to download your ASDFs and FITS to the same place.

Cadair:
Fixes #171

### DIFF
--- a/dkist/io/file_manager.py
+++ b/dkist/io/file_manager.py
@@ -33,6 +33,7 @@ from dkist.io.dask_utils import stack_loader_array
 from dkist.io.loaders import BaseFITSLoader
 from dkist.net import conf as net_conf
 from dkist.net.helpers import _orchestrate_transfer_task
+from dkist.utils.inventory import humanize_inventory
 
 
 class BaseStripedExternalArray:
@@ -328,6 +329,11 @@ class FileManager(BaseFileManager):
             ``ds.meta['inventory']``. An example of this would be
             ``path="~/dkist/{datasetId}"`` to save the files in a folder named
             with the dataset ID being downloaded.
+            If ``path`` is specified, and ``destination_endpoint`` is `None`
+            (i.e. you are downloading to a local Globus personal endpoint) this
+            method will set ``.basepath`` to the value of the ``path=``
+            argument, so that the array can be read from your transferred
+            files.
 
         destination_endpoint : `str`, optional
             A unique specifier for a Globus endpoint. If `None` a local
@@ -352,11 +358,12 @@ class FileManager(BaseFileManager):
             )
 
         inv = self._ndcube.meta["inventory"]
+        human_inv = humanize_inventory(inv)
 
         base_path = Path(net_conf.dataset_path.format(**inv))
         destination_path = path or self.basepath or "/~/"
         destination_path = Path(destination_path).as_posix()
-        destination_path = Path(destination_path.format(**inv))
+        destination_path = Path(destination_path.format(**human_inv))
 
         # TODO: If we are transferring the whole dataset then we should use the
         # directory not the list of all the files in it.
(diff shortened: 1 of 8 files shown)


dkist/io/file_manager.py
```diff
@@ -352,11 +358,12 @@ def download(self, path=None, destination_endpoint=None, progress=True, wait=Tru
             )
 
         inv = self._ndcube.meta["inventory"]
+        human_inv = humanize_inventory(inv)
 
         base_path = Path(net_conf.dataset_path.format(**inv))
         destination_path = path or self.basepath or "/~/"
         destination_path = Path(destination_path).as_posix()
-        destination_path = Path(destination_path.format(**inv))
+        destination_path = Path(destination_path.format(**human_inv))
```
eigenbrot, line 366: Not strictly related to this PR, per se, but it might be nice to have some custom error handling if a user tries to interpolate a key that doesn't exist; right now I'm fairly certain it would just throw `KeyError: 'foo'`. Even something like a quick note on how to show available keys would probably help a lot.

Cadair: If there's a new key in the inventory but it's not in the map it just copies it through without translating it, if that's what you mean.

dkist/utils/inventory.py
```diff
@@ -0,0 +1,82 @@
...
+
+class DefaultMap(defaultdict):
+    """
+    A tweak of default dict where the default value is the key that's missing.
+    """
+    def __missing__(self, key):
+        return key
+
+
+INVENTORY_KEY_MAP: Dict[str, str] = DefaultMap(None, {
```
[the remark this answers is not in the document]
Cadair: This file counts as docs right?!

dkist/io/file_manager.py
```diff
@@ -328,6 +329,11 @@ def download(self, path=None, destination_endpoint=None, progress=True, wait=Tru
             ``ds.meta['inventory']``. An example of this would be
             ``path="~/dkist/{datasetId}"`` to save the files in a folder named
             with the dataset ID being downloaded.
+            If ``path`` is specified, and ``destination_endpoint`` is `None`
+            (i.e. you are downloading to a local Globus personal endpoint) this
+            method will set ``.basepath`` to the value of the ``path=``
+            argument, so that the array can be read from your transferred
+            files.
```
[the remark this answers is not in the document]
Cadair: There is another use case, which is where you use this API to orchestrate a transfer to a Globus endpoint which isn't accessible from this process. In that case after the transfer you still wont be able to load the data.

dkist/io/file_manager.py
eigenbrot: Interesting. I think I see. Are you saving if a user has `destination_path = "{datasetID}_{somekey}_data"` then it will interpolate to, e.g., "YAFMS_{somekey}_data"? If so then I'm fine with that.

Cadair: oh no, I was mistaken. You can only use keys which exist.

For Fido there's a list, here there isn't really a way. I think I will open an issue to track that.

dkist/utils/inventory.py
```diff
@@ -0,0 +1,82 @@
...
+    Convert an inventory dict to have human readable keys.
+    """
+    key_converter = DefaultMap(None, INVENTORY_KEY_MAP)
+    humanized_inventory = {}
+    for key, value in inventory.items():
+        humanized_inventory[key_converter[key]] = value
+    return humanized_inventory
+
+
+def dehumanize_inventory(humanized_inventory: Dict[str, str]) -> Dict[str, str]:
```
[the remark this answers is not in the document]
Cadair: Fixed.

Cadair:
I am going to merge this with the CI busted and then fix it in a separate PR.

PR #182, opened 2022-07-18
Add a function to swap over axes of a dataset

Cadair:
This should help people to work with the data where the VISP has transposed the axes.

The general flow should be something like this:

```python

import dkist
from dkist.processing.caveats import swap_world_axes


ds = dkist.Dataset.from_asdf("path/to/file.asdf")
new_ds = swap_world_axes(ds, 1, 2)
```

I am going to try this out on an actual VISP dataset shortly


Cadair:
This approach is a dead end.

PR #197, opened 2022-08-12
Fix reauth

### LINKED ISSUE #179: When authing with Globus the first request after successful auth fails
Have no auth cache, run something which needs auth, see error, rerun command it works fine.

eigenbrot:
Stale tokens were not triggering a re-authentication request due to a change in how the message was delivered from Globus. This fixes that (and fixes #179 )


Cadair:
I think this needs a rebase on `main`.

Cadair (approved):

PR #201, opened 2022-08-23
Tidy `net.attrs` namespace

SolarDrew:
Adds a layer of indirection to import just the names we want available into the `dkist.net.attrs`/`a.dkist` namespace. This is the same approach used by sunpy.


SolarDrew:
CLoses #188

Cadair:
pre-commit.ci autofix

dkist/net/_attrs.py
```diff
@@ -0,0 +1,334 @@
+"""
+Search attributes which are specific to the `dkist.net.DKISTClient`.
+
+Other attributes provided by `sunpy.net.attrs` are supported by the client.
+"""
+import astropy.units as u
+import sunpy.net._attrs as _sunpy_attrs
+from sunpy.coordinates.frames import Helioprojective
+from sunpy.coordinates.utils import get_rectangle_coordinates
```
Cadair, line 9: This might be a little overkill.

I think we might just get away with this in the original `attrs.py` file?

```suggestion
import astropy.units as _u
import sunpy.net._attrs as _sunpy_attrs
from sunpy.coordinates.frames import _Helioprojective
from sunpy.coordinates.utils import _get_rectangle_coordinates
```

SolarDrew: I suppose so, but is that preferable? I assume sunpy has chosen to do it this way for a reason?

SolarDrew: Makes sense. Stand by for more commits then.

Cadair (approved):

PR #204, opened 2022-09-16
Manage empty searches in `transfer_complete_datasets`

### LINKED ISSUE #194: `transfer_complete_datasets` fails when given string dataset ids as input
### Description

`transfer_complete_datasets()` doesn't work with string input

### Expected vs Actual behavior

`transfer_complete_datasets()` should download a full dataset. It dies instead.

```
In [50]: transfer_complete_datasets('ABCDE')
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Input In [50], in <cell line: 1>()
----> 1 transfer_complete_datasets('ABCDE')

File ~/oss-projects/dkist/dkist/net/helpers.py:73, in transfer_complete_datasets(datasets, path, destination_endpoint, progress, wait, label)
     70 from dkist.net import conf
     72 if isinstance(datasets, str):
---> 73     datasets = _get_dataset_inventory(datasets)[0]
     75 # If we have a UnifiedResponse object, it could contain one or more dkist tables.
     76 # Stack them and then treat them like we were passed a single table with many rows.
     77 if isinstance(datasets, UnifiedResponse):

File ~/anaconda3/envs/dkist/lib/python3.10/site-packages/astropy/table/table.py:1874, in Table.__getitem__(self, item)
   1872     return self.columns[item]
   1873 elif isinstance(item, (int, np.integer)):
-> 1874     return self.Row(self, item)
   1875 elif (isinstance(item, np.ndarray) and item.shape == () and item.dtype.kind == 'i'):
   1876     return self.Row(self, item.item())

File ~/anaconda3/envs/dkist/lib/python3.10/site-packages/astropy/table/row.py:39, in Row.__init__(self, table, index)
     36 n = len(table)
     38 if index < -n or index >= n:
---> 39     raise IndexError('index {} out of range for table with length {}'
     40                      .format(index, len(table)))
     42 # Finally, ensure the index is positive [#8422] and set Row attributes
     43 self._index = index % n

IndexError: index 0 out of range for table with length 0
```

### Steps to Reproduce

```transfer_complete_datasets('ABCDE')```

### PR DESCRIPTION
**Pull request by SolarDrew**
Closes #194

In `transfer_complete_datasets`, if `_get_dataset_inventory` returns no results (which it might if the ID string is invalid or if the data just aren't available), this catches the error and raises it with an appropriate message.

### DIFF
--- a/dkist/net/helpers.py
+++ b/dkist/net/helpers.py
@@ -70,7 +70,10 @@ def transfer_complete_datasets(datasets: Union[str, QueryResponseRow, DKISTQuery
     from dkist.net import conf
 
     if isinstance(datasets, str):
-        datasets = _get_dataset_inventory(datasets)[0]
+        try:
+            datasets = _get_dataset_inventory(datasets)[0]
+        except IndexError as e:
+            raise IndexError(f"No results available for dataset {datasets}") from e
 
     # If we have a UnifiedResponse object, it could contain one or more dkist tables.
     # Stack them and then treat them like we were passed a single table with many rows.
--- a/dkist/net/helpers.py
+++ b/dkist/net/helpers.py
@@ -22,7 +22,7 @@ def _get_dataset_inventory(dataset_id: str):  # pragma: no cover
     Do a search for a single dataset id
     """
     results = DKISTClient().search(Dataset(dataset_id))
-    if len(results) == 1:
+    if len(results) == 0:
         raise IndexError(f"No results available for dataset {dataset_id}")
 
     return results
--- a/dkist/net/helpers.py
+++ b/dkist/net/helpers.py
@@ -21,7 +21,11 @@ def _get_dataset_inventory(dataset_id: str):  # pragma: no cover
     """
     Do a search for a single dataset id
     """
-    return DKISTClient().search(Dataset(dataset_id))
+    results = DKISTClient().search(Dataset(dataset_id))
+    if len(results) == 1:
+        raise IndexError(f"No results available for dataset {dataset_id}")
+
+    return results
 
 
 def transfer_complete_datasets(datasets: Union[str, QueryResponseRow, DKISTQueryResponseTable, UnifiedResponse],
@@ -70,10 +74,7 @@ def transfer_complete_datasets(datasets: Union[str, QueryResponseRow, DKISTQuery
     from dkist.net import conf
 
     if isinstance(datasets, str):
-        try:
-            datasets = _get_dataset_inventory(datasets)[0]
-        except IndexError as e:
-            raise IndexError(f"No results available for dataset {datasets}") from e
+        datasets = _get_dataset_inventory(datasets)[0]
 
     # If we have a UnifiedResponse object, it could contain one or more dkist tables.
     # Stack them and then treat them like we were passed a single table with many rows.
--- a/dkist/net/helpers.py
+++ b/dkist/net/helpers.py
@@ -23,7 +23,7 @@ def _get_dataset_inventory(dataset_id: str):  # pragma: no cover
     """
     results = DKISTClient().search(Dataset(dataset_id))
     if len(results) == 0:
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: dkist/net/helpers.py [hunk 1]
```diff
@@ -70,7 +70,10 @@ def transfer_complete_datasets(datasets: Union[str, QueryResponseRow, DKISTQuery
     from dkist.net import conf
 
     if isinstance(datasets, str):
-        datasets = _get_dataset_inventory(datasets)[0]
+        try:
+            datasets = _get_dataset_inventory(datasets)[0]
+        except IndexError as e:
+            raise IndexError(f"No results available for dataset {datasets}") from e
```
Cadair, line 76: If we do this inside `_get_dataset_inventory` we could avoid the `try` block which is always my slight preference?

dkist/net/helpers.py
```diff
@@ -21,7 +21,11 @@ def _get_dataset_inventory(dataset_id: str):  # pragma: no cover
     """
     Do a search for a single dataset id
     """
-    return DKISTClient().search(Dataset(dataset_id))
+    results = DKISTClient().search(Dataset(dataset_id))
+    if len(results) == 1:
```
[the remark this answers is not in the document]
SolarDrew: :facepalm: Yes, of course it should.

dkist/net/helpers.py
```diff
@@ -21,7 +21,11 @@ def _get_dataset_inventory(dataset_id: str):  # pragma: no cover
     """
     Do a search for a single dataset id
     """
-    return DKISTClient().search(Dataset(dataset_id))
+    results = DKISTClient().search(Dataset(dataset_id))
+    if len(results) == 0:
+        raise IndexError(f"No results available for dataset {dataset_id}")
```
Cadair, line 26:
```suggestion
        raise ValueError(f"No results available for dataset {dataset_id}")
```

dkist/net/tests/test_helpers.py
```diff
@@ -44,6 +44,17 @@ def test_download_default_keywords(orchestrate_transfer_mock, keywords):
     )
 
 
+def test_transfer_unavailable_data(mocker):
+    get_inv_mock = mocker.patch(
+        "dkist.net.client.DKISTClient.search",
```
Cadair, line 49: Is this mocker actually being hit? I would have thought this should be

```suggestion
        "dkist.net.helpers.DKISTClient.search",
```

SolarDrew: If it wasn't, surely the test would hit the "An attempt was made to connect to the internet" error (which is what happens if I take the mocker out entirely).

SolarDrew:
Also it looks like several of the tests are fouling on a `MatplotlibDeprecationWarning`, but not in tests that I've touched at all so I don't really know why that's happening or how to fix it.

dkist/net/tests/test_helpers.py
```diff
@@ -44,6 +44,17 @@ def test_download_default_keywords(orchestrate_transfer_mock, keywords):
...
+        autospec=True,
+        return_value=[],
+    )
+
+    with pytest.raises(ValueError):
```
Cadair, line 54:
```suggestion
    with pytest.raises(ValueError, "No results available for dataset"):
```

Cadair (approved):

dkist/net/tests/test_helpers.py
```diff
@@ -44,6 +44,17 @@ def test_download_default_keywords(orchestrate_transfer_mock, keywords):
...
 
 
+def test_transfer_unavailable_data(mocker):
+    get_inv_mock = mocker.patch(
+        "dkist.net.client.DKISTClient.search",
+        autospec=True,
+        return_value=[],
+    )
+
+    with pytest.raises(ValueError, "No results available for dataset"):
```
Cadair, line 54:
```suggestion
    with pytest.raises(ValueError, match="No results available for dataset"):
```

PR #207, opened 2022-09-26
Provide documentation of what path interpolation parameters you can use in Globus download methods

### LINKED ISSUE #180: Provide documentation of what path interpolation parameters you can use in Globus download methods
See https://github.com/DKISTDC/dkist/pull/178/files#r912967338

### LINKED ISSUE #185: `Fido.fetch` path interpolation keys are not the same as `FileManager.download`
The sunpy machinery for path interpolation lowercases and adds underscores which isn't happening in the non-sunpy UT globus code.

SolarDrew:
Closes #180 
Fixes #185 

Not sure if this is the best place in the docs structure to have this information, but it's easily moved. I also can't remember what definition of dataset DKIST uses so the given example may be nonsense but I can also change that.


docs/guide/net.rst
```diff
@@ -183,3 +186,61 @@ Finally, if you know the dataset ID of a dataset you wish to download, you can j
...
+
+Path interpolation keys
+-----------------------
+
+When downloading DKIST data with ``ds.files.download()``, the ``path=`` keyword argument can be used to specify the target folder for the download.
+This path can include keys corresponding to metadata entries, and those values are then used to complete the download path.
+For example, to download a dataset into its own folder named with the dataset ID, with separate subfolders for each instrument in the dataset, you could set ``path="~/data/dkist/{Dataset ID}/{Instrument}/"``.
+This would take the values for the ``datasetID`` and ``instrumentName`` metadata keywords in the dataset and insert them into the path string.
+
+Here is a full list of the metadata keywords available for this purpose and their corresponding path interpolation keys:
```
Cadair, line 200: Can we auto generate this? There will soon be more keys in here and having to *remember* to update it everytime seems like a bad idea. If we have a Python API for printing it we could call it in a code block in the docs?

SolarDrew: Yes that's a great call, I'll get on it.

docs/guide/net.rst
```diff
@@ -124,6 +124,9 @@ Once we have loaded the dataset, if we wish to transfer all the FITS files a sin
 If this is the first time you have run this method, or your authentication has expired, a login page should open in your webbrowser to authenticate with the Globus system.
 By default this call will download all the FITS files to the current value of ``ds.files.basepath``, i.e. by default in the same directory as the loaded ASDF file.
 You can override this behaviour by using the ``path=`` keyword argument.
+The path argument can contain keys which will be replaced by the corresponding values in the dataset's metadata.
+For example, setting `path="~/data/dkist/{Instrument}/"` will download all files and save them in separate folders named for the value of the `instrumentName` keyword in the header.
```
Cadair, line 128:
```suggestion
For example, setting `path="~/data/dkist/{instrument}/"` will download all files and save them in separate folders named for the instrument.
```

docs/guide/net.rst
```diff
@@ -183,3 +186,19 @@ Finally, if you know the dataset ID of a dataset you wish to download, you can j
 .. code-block:: python
 
     transfer_complete_datasets("AAAAA")
+
+.. _interpolation-keys:
+
+Path interpolation keys
+-----------------------
+
+When downloading DKIST data with ``ds.files.download()``, the ``path=`` keyword argument can be used to specify the target folder for the download.
```
Cadair, line 195:
```suggestion
When downloading DKIST data with ``ds.files.download()`` or ``Fido.fetch(), the ``path=`` keyword argument can be used to specify the target folder for the download.
```

docs/guide/net.rst
```diff
@@ -183,3 +186,19 @@ Finally, if you know the dataset ID of a dataset you wish to download, you can j
...
+This path can include keys corresponding to metadata entries, and those values are then used to complete the download path.
+For example, to download a dataset into its own folder named with the dataset ID, with separate subfolders for each instrument in the dataset, you could set ``path="~/data/dkist/{Dataset ID}/{Instrument}/"``.
+This would take the values for the ``datasetID`` and ``instrumentName`` metadata keywords in the dataset and insert them into the path string.
```
Cadair, line 198:
```suggestion
For example, to download a dataset into its own folder named with the dataset ID, with separate subfolders for each instrument in the dataset, you could set ``path="~/data/dkist/{dataset_id}/{instrument}/"``.
This would take the values for the dataset ID and instrument name from either the ASDF file or the search results.
```

dkist/utils/inventory.py
```diff
@@ -60,19 +64,24 @@ def __missing__(self, key):
 })
 
 
+def _key_clean(key):
```
SolarDrew, line 67: I thought we wanted to avoid having to define this here?

Cadair: Yeah, we did, but my little idea of making a dict full of `None`s didn't work.

Cadair (approved):

PR #213, opened 2022-12-13
Make Dataset.from_asdf work with TiledDataset

Cadair:
yes it's weird but here we are.


eigenbrot (approved):
I'm not sure I get what's going on here, but the code looks good!

PR #217, opened 2023-01-05
Fix CI

Cadair:


Cadair:
The devdeps is broken because of https://github.com/sunpy/ndcube/pull/584 which means it probably isn't on us to fix.

PR #221, opened 2023-02-10
Require that all Datasets have units

Cadair:
We have not so far been writing the unit for the main array into the ASDF file, this PR makes a new dataset ASDF schema which makes unit required, as well as fixing the schema for the optional unit property in the 1.0.0 schema.


dkist/io/asdf/resources/schemas/tiled_dataset-1.0.0.yaml
```diff
@@ -16,7 +16,9 @@ properties:
     items:
       type: array
       items:
-        tag: "asdf://dkist.nso.edu/tags/dataset-1.0.0"
+        anyOf:
+          - tag: "asdf://dkist.nso.edu/tags/dataset-1.0.0"
+          - tag: "asdf://dkist.nso.edu/tags/dataset-1.1.0"
```
WilliamJamieson, line 21: This might require a new version of this schema in the manifest alongside dataset-1.1.0. It should be tested to make sure old and new files work here. (This is probably fine, but it could be a pain point)

Cadair: Why would that be the case?

Cadair:
Before we merge this we need tests of all the versions of the tiled_dataset schema.

WilliamJamieson:
> It seems that the addition of the 1.1.0 schema isn't being used on write.

I think you need to add the "new" manifest extension: like https://github.com/DKISTDC/dkist/blob/7ec32ce3eb216d0be10366dc07ee40dd42811052/dkist/io/asdf/entry_points.py#L47-L48 only with `dkist-1.1.0`?

dkist/io/asdf/resources/schemas/dataset-1.0.0.yaml
```diff
@@ -23,7 +23,9 @@ properties:
     tag: "core/ndarray-1.0.0"
 
   unit:
-    tag: "unit/unit-1.0.0"
+    anyOf:
+      - tag: "unit/unit-1.0.0"
+      - tag: "tag:stsci.edu:asdf/unit/unit-1.0.0"
```
eigenbrot, line 28: Just to check my understanding; this change to the 1.0.0 schema is to allow these files to be correctly read by the new unit machinery if units are present, but to not require that they have it?...

dkist/io/asdf/resources/schemas/dataset-1.1.0.yaml
```diff
@@ -0,0 +1,49 @@
...
+    description: The coordinate system for the complete dataset.
+    anyOf:
+      - tag: "tag:stsci.edu:gwcs/wcs-1.0.0"
+      - tag: "tag:stsci.edu:gwcs/wcs-1.1.0"
+
+  mask:
+    tag: "tag:stsci.edu:asdf/core/ndarray-1.0.0"
+
+  unit:
+    tag: "tag:stsci.edu:asdf/unit/unit-1.0.0"
```
eigenbrot, line 26: ....whereas down here we require that all 1.1.0 datasets have units. Is that right?

dkist/io/asdf/tests/test_dataset.py
```diff
@@ -111,17 +111,17 @@ def test_save_dataset_with_file_schema(tagobj, tmpdir):
...
+def test_read_all_schema_versions(eit_dataset_asdf_path):
     """
     This test validates that we can successfully read a full and valid Dataset
     object from files with all versions of the dataset schema.
     """
     with importlib_resources.as_file(importlib_resources.files("dkist.io") / "level_1_dataset_schema.yaml") as schema_path:
-        with asdf.open(asdf_file) as afile:
+        # Firstly verify that the tag versions in the test filename are the ones used in the file
+        with asdf.open(eit_dataset_asdf_path, _force_raw_types=True) as afile:
+            assert afile["dataset"]._tag.rsplit("/")[-1] in str(eit_dataset_asdf_path)
```
eigenbrot, line 122: Relying on the path to a test file seems a bit fragile, but w/e.

dkist/tests/generate_eit_test_dataset.py
```diff
@@ -7,12 +7,13 @@
...
 from astropy.modeling.models import (AffineTransformation2D, Multiply,
                                      Pix2Sky_TAN, RotateNative2Celestial, Shift)
+from astropy.table import Table
 from astropy.time import Time
 from gwcs import coordinate_frames as cf
 from sunpy.time import parse_time
 
 from dkist_inventory.asdf_generator import references_from_filenames
-from dkist_inventory.inventory import table_from_headers
+from dkist_inventory.header_parsing import HeaderParser
```
eigenbrot, line 16: Strictly speaking, I'd prefer to see changes needed for the new version of `dkist-inventory` in their own separate PR. It's not a hill I'm going to die on, though.

dkist/tests/generate_eit_tiled_dataset.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+mosaic = MosaicedVBIBlueDataset(2, 1, linewave=500*u.nm, detector_shape=(2, 2))
+
+with tempfile.TemporaryDirectory() as tempdir:
+    mosaic.generate_files(tempdir, required_only=True)
+
+    asdf_filename = rootdir / "test_tiled_dataset.asdf"
+    ds = dataset_from_fits(tempdir, asdf_filename)
+    print(ds)
```
eigenbrot, line 26: Debug cruft?

eigenbrot (approved):
My only hope is that I understood what is going on here just barely enough to make this approval mean something.

dkist/io/asdf/resources/schemas/dataset-1.0.0.yaml
Cadair: This fixes a bug in the 1.0.0 schema where it wouldn't validate the unit. As it's a backwards compatible change we can make it without updating the schema version.

dkist/io/asdf/resources/schemas/dataset-1.1.0.yaml
Cadair: yes, which is a backwards-incompatible change, so a new schema version is required.

dkist/io/asdf/tests/test_dataset.py
Cadair: Yeah, the alternative was a lot more work!

dkist/tests/generate_eit_test_dataset.py
Cadair: Both these files are just to generate the test data when we update the asdf schemas, so I have no issues about them being hacky and tied to whatever version of inventory I like :grinning:

dkist/tests/generate_eit_tiled_dataset.py
Cadair: legitimate stdout from a script :)

PR #225, opened 2023-02-18
Fix lon_pole units in VaryingCelestialTransform

nabobalis:
I added a simple test. 

Not sure if there is something else to do.


dkist/wcs/tests/test_models.py
```diff
@@ -46,6 +46,19 @@ def test_generate_celestial_unitless():
     assert u.allclose(shift1.offset, 0)
 
 
+def test_varying_transform_no_lon_pole_unit():
```
nabobalis, line 49: If you prefer, I can remove this test and add the unit check to the test below and remove the lon_pole argument in the `VaryingCelestialTransform` call.

I can also add a test for `generate_celestial_transform` instead? Or both?

.github/workflows/main.yml
```diff
@@ -4,21 +4,28 @@ on:
...
-        - macos: py311
-        - linux: py310-devdeps
+        - linux: py311
+        - windows: py310
+        - macos: py39
+        - macos: py310-devdeps
         - windows: py38-oldestdeps
-      coverage: 'codecov'
+        - linux: py311-build_docs
+          pytest: false
```
nabobalis, line 23: This did not work even though I thought it would.

setup.cfg
```diff
@@ -52,9 +53,10 @@ docs =
   sphinx-gallery
   sphinx-changelog
   pytest
-  towncrier
   sphinx_autodoc_typehints
   dkist-sphinx-theme>=1.0.0rc6
+  # Not a direct dependency, but we need to pin this until sphinx-changelog is updated (see sunpy #6668)
+  towncrier<22.12.0
```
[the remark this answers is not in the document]
nabobalis: Fixed.

nabobalis:
`sunpy:fido_guide` is not found. I assume due to us removing it/renaming it?

Cadair:
> `sunpy:fido_guide` is not found. I assume due to us removing it/renaming it?

I re-pointed our intersphinx to stable, but maybe we can fix this before release.

Cadair:
Thanks a lot @nabobalis :partying_face:

PR #232, opened 2023-03-24
Add the ability to set chunksize when making the dask array

Cadair:
This PR got a little :ox: shaving, I removed the use of `ExternalArrayReference`, however, its primary purpose is to provide the primatives to allow the optional specification of a chunksize when making a dask array from each loader referenced by an asdf.

This should be all the changes needed to test the performance of loading a chunked array, and then changes to the asdf maker can use them later.

This also cleans up the implementation of the Loader class by making the abstract method the `__getitem__` method, this is needed to access `.section` and read only a subset of the file.

TODO:

- [x] Write a test jig which loads an asdf file and then retrospectively injects a chunksize and benchmark it.


dkist/io/loaders.py
```diff
@@ -13,48 +14,61 @@
...
-    def __array__(self):
-        return self.fits_array
+    @property
+    def data(self):
+        return self[:]
 
+    @abc.abstractmethod
     def __getitem__(self, slc):
-        return self.fits_array[slc]
+        pass
```
SolarDrew, line 71: Should this method be defined if it's not doing anything?

SolarDrew (commented):
LGTM :+1:

dkist/io/loaders.py
SolarDrew: Never mind, it's an abstract method and that's fine. Carry on.

SolarDrew (approved):

PR #248, opened 2023-06-02
Update Ravel to support N dimensional data.

svensk-pojke:


dkist/io/asdf/converters/models.py
```diff
@@ -123,7 +123,7 @@ class RavelConverter(TransformConverterBase):
     """
 
     tags  = [
-        "asdf://dkist.nso.edu/tags/ravel_model-1.0.0"
+        "asdf://dkist.nso.edu/tags/ravel_model-1.1.0"
```
Cadair, line 126: undo this and your test issue should go away

PR #249, opened 2023-06-02
Update Ravel to support N-dimensional data

svensk-pojke:
Modifies Ravel to use arbitrary multi-dimensional lookup tables. This also include updating Ravel and Unravel to support units if they are present. The latter change supersedes the Ravel with units PR.


dkist/wcs/models.py
```diff
@@ -739,12 +739,22 @@ def varying_celestial_transform_from_tables(crpix: Union[Iterable[float], u.Quan
...
+    def input_units(self):
+        return {'x': u.pix}
+
+    @property
+    def return_units(self):
+        return {'y': u.pix}
 
     def __init__(self, array_shape, order='C', **kwargs):
+        # n_inputs is a class attribute, but it is unknown until run-time
+        self.__class__.n_inputs = len(array_shape)
```
Cadair, line 757: This is a deal breaker. If you had two Ravel models with different dimensions going at the same time this wouldn't work. For example I would expect this to error

```python
m = Ravel((10, 10, 10)) & Ravel((10, 10))
m(0, 1, 2, 3, 4)
```

because you are setting the variable on the Ravel class itself not the instance so both instances of that Ravel model would have `n_inputs = 2` as that would be the last one.

Would this not work if you make it a property? i.e.
```python
@property
def  n_inputs(self):
    return len(self.array_shape)
```

dkist/wcs/models.py
```diff
@@ -753,27 +763,56 @@ def __init__(self, array_shape, order='C', **kwargs):
...
+        if hasattr(inputs[0], "unit"):
+            input_units = inputs[0].unit
+            input_values = [item[0].value for item in inputs]
+        else:
+            input_units = None
+            input_values = [item[0] for item in inputs]
+        # round the index values, but clip them if they exceed the array bounds
+        # the bounds are one lass than the shape dimension value
+        array_bounds = tuple(np.array(self.array_shape) - 1)
+        # array_bounds = [item - 1 for item in self.array_shape]
```
Cadair, line 777: Should this comment get deleted?

```suggestion
```

dkist/wcs/models.py
```diff
@@ -753,27 +763,56 @@ def __init__(self, array_shape, order='C', **kwargs):
...
+        rounded_inputs = np.clip(np.rint(input_values).astype(int), None, array_bounds)
+        result = np.ravel_multi_index(rounded_inputs, self.array_shape, order=self.order).astype(float)
```
Cadair, line 779: Why the `astype(float)` here?

dkist/wcs/models.py
```diff
@@ -753,27 +763,56 @@ def __init__(self, array_shape, order='C', **kwargs):
...
+        index = 0 if self.order == "F" else -1
+        # Adjust the result to allow a fractional part for interpolation in Tabular1D
+        fraction = input_values[index] - rounded_inputs[index]
+        result += fraction
+        # Put the units back if they were there...
+        if input_units is not None:
+            result = result * input_units
+            # result = np.array([u.Quantity(item, unit=input_units) for item in result])
```
Cadair, line 787: Also delete this comment?
```suggestion
```

dkist/wcs/models.py
```diff
@@ -753,27 +763,56 @@ def __init__(self, array_shape, order='C', **kwargs):
...
+    @property
+    def input_units(self):
+        return {'x': u.pix}
+
+    @property
+    def return_units(self):
+        return {'y': u.pix}
 
     def __init__(self, array_shape, order='C', **kwargs):
+        self.__class__.n_outputs = len(array_shape)
```
Cadair, line 815: Same here.

dkist/wcs/models.py
```diff
@@ -782,12 +821,27 @@ def __init__(self, array_shape, order='C', **kwargs):
...
+            input_value = input_[0]
+        result = list(np.unravel_index(int(input_value), self.array_shape, order=self.order))
+        result = [np.array([float(item)]) for item in result]
+        # Adjust the result to allow a fractional part for interpolation in Tabular1D
+        index = 0 if self.order == "F" else -1
+        fraction = input_value % 1
+        result[index] += fraction
+        if input_unit is not None:
+            result = result * input_unit
+            # result = np.array([u.Quantity(item, unit=input_unit) for item in result])
```
Cadair, line 838:
```suggestion
```

dkist/wcs/models.py
```diff
@@ -753,27 +763,56 @@ def __init__(self, array_shape, order='C', **kwargs):
...
-        ravel_shape = self.array_shape[1]
-        if self.order == 'F':
-            inputs = inputs[::-1]
-            ravel_shape = self.array_shape[0]
-        return (np.round(inputs[0]) * ravel_shape) + inputs[1]
+        # If we have units, then all values must have units, hence test for existence only on the first one.
+        # Also, assume all units are the same.
+        if hasattr(inputs[0], "unit"):
+            input_units = inputs[0].unit
+            input_values = [item[0].value for item in inputs]
```
Cadair, line 770: In general I always recommend using `to_value` as it's more explicit.
```suggestion
            input_values = [item[0].to_value(u.pix) for item in inputs]
```

dkist/wcs/models.py
Cadair, line 779: Do we have tests for array inputs to `Ravel`, i.e. if we tried to compute the world coordinates for all pixel coordinates at once?

dkist/wcs/models.py
```diff
@@ -753,27 +763,56 @@ def __init__(self, array_shape, order='C', **kwargs):
...
+            input_units = None
+            input_values = [item[0] for item in inputs]
+        # round the index values, but clip them if they exceed the array bounds
+        # the bounds are one lass than the shape dimension value
+        array_bounds = tuple(np.array(self.array_shape) - 1)
+        # array_bounds = [item - 1 for item in self.array_shape]
+        rounded_inputs = np.clip(np.rint(input_values).astype(int), None, array_bounds)
+        result = np.ravel_multi_index(rounded_inputs, self.array_shape, order=self.order).astype(float)
+        index = 0 if self.order == "F" else -1
+        # Adjust the result to allow a fractional part for interpolation in Tabular1D
```
Cadair, line 781: Can you explain this bit to me, I am not sure it's obvious why it's done like this.

Cadair (changes requested):
This is awesome, thanks!

I have a bunch of inline comments, but my two main concerns are:

* The setting of `n_inputs`
* Testing of evaluation of arrays of inputs (which I am not sure we had before).

dkist/wcs/models.py
svensk-pojke: Done. This works if you defer the call to super().__init__(**kwargs) until after the instance attributes are set.

dkist/wcs/models.py
svensk-pojke: Done.

dkist/wcs/models.py
svensk-pojke: Because the input coordinates are floating point and we round all but the inner-most coordinate (inner in the sense of memory access) so that we can pass the interpolation on to Tabular1D.

dkist/wcs/models.py
svensk-pojke: Done.

dkist/wcs/models.py
svensk-pojke: Done.

dkist/wcs/models.py
svensk-pojke: Done.

dkist/wcs/models.py
svensk-pojke: Done.

dkist/wcs/models.py
svensk-pojke: Done, discussed in person.

dkist/wcs/models.py
```diff
@@ -739,55 +739,109 @@ def varying_celestial_transform_from_tables(crpix: Union[Iterable[float], u.Quan
...
-    n_inputs = 2
     n_outputs = 1
     _separable = False
+    _input_units_allow_dimensionless = True
 
-    def __init__(self, array_shape, order='C', **kwargs):
-        super().__init__(**kwargs)
+    @property
+    def n_inputs(self):
+        return len(self.array_shape)
```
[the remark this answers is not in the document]
svensk-pojke: Hi @WilliamJamieson, yes. We currently have a 3D case. The extension to 3D is generalizeable to N dimensions, so that is what I implemented, which allows 2D and 3D to be handled in the same code and allows for future functionality beyond 3D, if needed.

Cadair (approved):

PR #266, opened 2023-06-07
Various upgrades and fixes for `dkist.net`

### LINKED ISSUE #107: `import dkist` should not import `dkist.net`?
Pros: Don't have to import it again
Cons: Have to load all of Fido always

### LINKED ISSUE #229: New dataset inventory fields need search attrs
There are now some columns in the response we don't have friendly names for. We might also need search attrs for some of them.

### LINKED ISSUE #241: `dkist.net.transfer_complete_datasets()` should do path interpolation
I had assumed that `transfer_complete_datasets()` would be able to do path interpolation, a la Fido. For example, that you would be able to do `dkist.net.transfer_complete_datasets('AAAAA', path='/path/to/wherever/{instrument}/{dataset_id}')`. It turns out this isn't the case but it really would be useful and would bring the API into closer alignment with Fido.

### LINKED ISSUE #243: Searching dataset inventory by Wavelength is broken
Seem the dataset search API has changed to use a new `wavelengthRanges` parameter.

https://api.dkistdc.nso.edu/datasets/docs#/default/get_datasets_v1_get

### LINKED ISSUE #251: `a.dkist.Embargoed.false` doesn't work properly
### Description

`a.dkist.Embargoed.false` isn't evaulated properly and therefore returns invalid search results.

### Expected vs Actual behavior

Searching Fido with `a.dkist.Embargoed.false` should return results for which `Embargoed` is `False`. Instead it returns results for which `Embargoed` is `True`.

### Steps to Reproduce

```python
>>> res = Fido.search(a.dkist.Embargoed.false)

>>> len(res['dkist']['Embargoed'])
246

>>> len(res['dkist']['Embargoed'] == False)
246

```

### System Details

- DKIST Version: 1.0.0b13
- SunPy Version: 4.1.7
- Astropy Version: 5.3
- Python Version: 3.11.3
- OS information: Arch Linux

### LINKED ISSUE #261: `a.dkist.FriedParameter` doesn't accept units
Obviously it should. Assumes metres which is fine, but explicit is better than implicit.

SolarDrew:
Fixes #261 .
Fixes #243 .
Fixes #107 .
Fixes #251 .
Fixes #229.
Fixes #241.


dkist/net/attrs.py
```diff
@@ -157,6 +159,8 @@ class Embargoed(_SimpleAttr):
         A boolean determining if a dataset currently under embargo.
     """
     def __init__(self, is_embargoed: bool):
+        if isinstance(is_embargoed, str) and is_embargoed.lower() == "false":
+            is_embargoed = False
```
Cadair, line 163: What about `"true"` :wink: 

More over I think that if it's not `bool(True)` then it should be False. i.e. why is `a.Embargoed("wibble")` the same as True?

dkist/io/file_manager.py
```diff
@@ -419,6 +417,9 @@ def download(self, path=None, destination_endpoint=None, progress=True, wait=Tru
         label : `str`
             Label for the Globus transfer. If `None` then a default will be used.
         """
+        from dkist.net import conf as net_conf
```
Cadair, line 420:
```suggestion
        # Import here to prevent triggering an import of `.net` with `dkist.dataset`.
        from dkist.net import conf as net_conf
```

changelog/266.bugfix.2.rst
```diff
@@ -0,0 +1 @@
+Fix `a.dkist.Embargoed.false`.
```
Cadair, line 1:
```suggestion
Fix using `a.dkist.Embargoed.false` and `a.dkist.Embargoed.true` to specify embargo status.
```

dkist/net/attrs.py
```diff
@@ -356,3 +360,84 @@ def __init__(self, temporalmin: _u.s, temporalmax: _u.s):
 
     def collides(self, other):
         return isinstance(other, self.__class__)
+
+
+# highLevelSoftwareVersion
+class SoftwareVersion(_SimpleAttr):
```
Cadair, line 366: Maybe call this `SummitSoftwareVersion`?

dkist/net/attrs.py
```diff
@@ -356,3 +360,84 @@ def __init__(self, temporalmin: _u.s, temporalmax: _u.s):
...
+    ----------
+    workflow : `str`
+        Version of the workflow.
+    """
+    def __init__(self, workflow: str):
+        super().__init__(workflow)
+
+
+# observingProgramExecutionId
+class ObservingProgram(_SimpleAttr):
```
Cadair, line 408: I think add the `ExecutionID` parts to these, only real advanced users will care, so I think explicit is better.

Cadair (approved):

PR #271, opened 2023-06-23
Add header-slicing functionality to dataset

### LINKED ISSUE #134: Link the header table to the files / FileManager
When you slice the Dataset, it should (automatically?) be possible to view the subset of the headers table.
It should also be possible to look in the headers table and work out what index of the FileManager a row corresponds to etc

SolarDrew:
Fixes #134 

<del>Not actually working at the moment but it's very close. Pushing now so the approach can be looked at while I'm fixing the details.</del>

### DIFF
--- a/changelog/271.bugfix.rst
+++ b/changelog/271.bugfix.rst
@@ -1,1 +1,1 @@
-Make `.Dataset` return the appropriately sliced header table when slicing data.
+Make `dkist.Dataset` return the appropriately sliced header table when slicing data.
--- a/changelog/271.bugfix.rst
+++ b/changelog/271.bugfix.rst
@@ -1,0 +1,1 @@
+Make `Dataset` return the appropriately sliced header table when slicing data.
--- a/changelog/271.bugfix.rst
+++ b/changelog/271.bugfix.rst
@@ -1,1 +1,1 @@
-Make `Dataset` return the appropriately sliced header table when slicing data.
+Make `.Dataset` return the appropriately sliced header table when slicing data.
--- a/dkist/dataset/dataset.py
+++ b/dkist/dataset/dataset.py
@@ -124,17 +124,25 @@ class Dataset(NDCube):
         sliced_dataset = super().__getitem__(item)
         if self._file_manager is not None:
             sliced_dataset._file_manager = self._file_manager._slice_by_cube(item)
-            self.headers = self._slice_headers(item)
+            sliced_dataset.headers = self._slice_headers(item)
         return sliced_dataset
 
     def _slice_headers(self, slice_):
-        file_idx = self.files._array_slice_to_loader_slice(slice_)
+        idx = self.files._array_slice_to_loader_slice(slice_)
+        if idx == (np.s_[:],):
+            return self.all_headers.copy()
+        file_idx = []
+        for i in idx:
+            if not isinstance(i, slice):
+                i = slice(i, i+1)
+            file_idx.append(i)
+        file_idx = tuple(file_idx)
         grid = np.mgrid[{tuple: file_idx, slice: (file_idx,)}[type(file_idx)]]
         file_idx = tuple(grid[i].ravel() for i in range(grid.shape[0]))
-        files_shape = (i for i in self.files.shape if i != 1)
-        flat_index = np.ravel_multi_index(file_idx[::-1], files_shape[::-1])
+        files_shape = [i for i in self.files.fileuri_array.shape if i != 1]
+        flat_idx = np.ravel_multi_index(file_idx[::-1], files_shape[::-1])
 
-        return self.headers[flat_idx]
+        return self.all_headers[flat_idx].copy()
 
     """
     Properties.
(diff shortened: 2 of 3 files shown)


SolarDrew:
I think @Cadair suggested moving this to `ndcube.__getitem__`? Happy to do so but for now it's here and working. Could perhaps do with testing at higher dimensions but I'm not sure how best to achieve that.

dkist/dataset/dataset.py
```diff
@@ -114,21 +115,41 @@ def __init__(self, data, wcs=None, uncertainty=None, mask=None, meta=None,
         if "inventory" not in meta:
             raise ValueError("The meta dict must contain the inventory record.")
 
+        self.headers = meta["headers"].copy()
+
```
Cadair, line 119: Get rid of this, let's just modify meta.
```suggestion
```

dkist/dataset/dataset.py
```diff
@@ -114,21 +115,41 @@ def __init__(self, data, wcs=None, uncertainty=None, mask=None, meta=None,
...
         super().__init__(data, wcs, uncertainty=uncertainty, mask=mask, meta=meta,
                          unit=unit, copy=copy)
 
     def __getitem__(self, item):
         sliced_dataset = super().__getitem__(item)
         if self._file_manager is not None:
             sliced_dataset._file_manager = self._file_manager._slice_by_cube(item)
+            sliced_dataset.headers = self._slice_headers(item)
```
Cadair, line 127:
```suggestion
            sliced_dataset.meta["headers"] = self._slice_headers(item)
```

dkist/dataset/dataset.py
```diff
@@ -114,21 +115,41 @@ def __init__(self, data, wcs=None, uncertainty=None, mask=None, meta=None,
...
         return sliced_dataset
 
+    def _slice_headers(self, slice_):
+        idx = self.files._array_slice_to_loader_slice(slice_)
+        if idx == (np.s_[:],):
+            return self.all_headers.copy()
```
Cadair, line 133: :thinking:  I am not sure if we need the copies or not. I will think about it.

dkist/dataset/dataset.py
```diff
@@ -114,21 +115,41 @@ def __init__(self, data, wcs=None, uncertainty=None, mask=None, meta=None,
...
+
+        return self.all_headers[flat_idx].copy()
+
     """
     Properties.
     """
 
     @property
-    def headers(self):
+    def all_headers(self):
```
Cadair, line 152: Revert this as well, I don't think we need a way to access "all_headers".

SolarDrew:
Well something's gone wrong because now it's modifying the original header table in place when you slice the dataset, which obviously breaks everything next time you try to slice it. Still trying to figure out how to get round that while keeping your suggestions mostly in place.

changelog/271.bugfix.rst
```diff
@@ -0,0 +1 @@
+Make `Dataset` return the appropriately sliced header table when slicing data.
```
Cadair, line 1:
```suggestion
Make `.Dataset` return the appropriately sliced header table when slicing data.
```

dkist/dataset/dataset.py
```diff
@@ -121,8 +122,27 @@ def __getitem__(self, item):
...
         if self._file_manager is not None:
             sliced_dataset._file_manager = self._file_manager._slice_by_cube(item)
+            sliced_dataset.meta = sliced_dataset.meta.copy()
+            sliced_dataset.meta["headers"] = self._slice_headers(item)
         return sliced_dataset
 
+    def _slice_headers(self, slice_):
+        idx = self.files._array_slice_to_loader_slice(slice_)
+        if idx == (np.s_[:],):
+            return self.headers.copy()
```
Cadair, line 132:
```suggestion
            return self.headers
```

dkist/dataset/tests/test_dataset.py
Cadair, line 1: We should add two more tests, one that checks to see if this works with a >1D file array and one which confirms that modifying the sliced dataset's header table changes the parent dataset's header table

dkist/dataset/dataset.py
```diff
@@ -121,8 +122,27 @@ def __getitem__(self, item):
         sliced_dataset = super().__getitem__(item)
         if self._file_manager is not None:
             sliced_dataset._file_manager = self._file_manager._slice_by_cube(item)
+            sliced_dataset.meta = sliced_dataset.meta.copy()
```
SolarDrew, line 125: This is stopping changes from the new, sliced, headers from being reflected in the table on the original dataset object. Not sure what to do about that, since we need it to make sure the original header table doesn't get sliced. How does file manager solve this problem?

dkist/dataset/dataset.py
```diff
@@ -121,8 +132,29 @@ def __getitem__(self, item):
...
+            sliced_dataset.meta["headers"] = self._slice_headers(item)
         return sliced_dataset
 
+    def _slice_headers(self, slice_):
+        idx = self.files._array_slice_to_loader_slice(slice_)
+        if idx == (np.s_[:],):
+            return self.headers
```
Cadair, line 142:
```suggestion
            return self.headers.copy()
```

changelog/271.bugfix.rst
```diff
@@ -0,0 +1 @@
+Make `.Dataset` return the appropriately sliced header table when slicing data.
```
Cadair, line 1: Apparently this shorthand was ambiguous
```suggestion
Make `dkist.Dataset` return the appropriately sliced header table when slicing data.
```

Cadair (approved):

PR #274, opened 2023-07-05
Move dataset loading methods out into a single-dispatch function

### LINKED ISSUE #128: Improve / change the dataset loading API
A couple of issues with the current API:

- Only supports one path at a time, meaning you can't pass a whole bunch of downloaded asdfs from `Fido.fetch` and get a bunch of datasets back.
- You use a class method on `Dataset` to get an instance of `TiledDataset` which is weird.

### LINKED ISSUE #276: jsonschema is not listed as a dependency
### Description

jsonschema is imported in dataset.py
https://github.com/DKISTDC/dkist/blob/d082995338e69dffd803ac0bdef4e8a535449d19/dkist/dataset/dataset.py#L6

but it not listed as a dependency in setup.cfg
https://github.com/DKISTDC/dkist/blob/d082995338e69dffd803ac0bdef4e8a535449d19/setup.cfg#L23-L36

This is and was previously provided via the dependency on asdf (which is currently evaluating options for dealing with breaking changes in jsonschema 4.18 including dropping jsonschema as a dependency).

One option would be to add jsonschema as a dependency. A second option would be to increase the lower asdf pin to 2.15 and import 'ValidationError' from [asdf.exceptions](https://github.com/asdf-format/asdf/blob/c3e358886a41c36b81a2a26bd03aedc2720c9817/asdf/exceptions.py#L9).

SolarDrew:
Fixes #128 
Fixes #276 

It's a bit messy at the moment, will want some tidying up.

### DIFF
--- a/changelog/274.feature.rst
+++ b/changelog/274.feature.rst
@@ -1,0 +1,1 @@
+Add a new `load_dataset()` function to combine and replace `Dataset.from_directory()` and `Dataset.from_asdf()`.
--- a/changelog/274.feature.rst
+++ b/changelog/274.feature.rst
@@ -1,1 +1,1 @@
-Add a new `load_dataset()` function to combine and replace `Dataset.from_directory()` and `Dataset.from_asdf()`.
+Add a new `dkist.load_dataset` function to combine and replace ``Dataset.from_directory()`` and ``Dataset.from_asdf()``.
--- a/dkist/dataset/__init__.py
+++ b/dkist/dataset/__init__.py
@@ -1,2 +1,3 @@
 from .dataset import Dataset
+from .loader import load_dataset
 from .tiled_dataset import TiledDataset
--- a/dkist/dataset/dataset.py
+++ b/dkist/dataset/dataset.py
@@ -1,10 +1,5 @@
-import importlib.resources as importlib_resources
-from pathlib import Path
 from textwrap import dedent
 
-from jsonschema.exceptions import ValidationError
-
-import asdf
 import gwcs
 from astropy.wcs.wcsapi.wrappers import SlicedLowLevelWCS
 from ndcube.ndcube import NDCube, NDCubeLinkedDescriptor
@@ -159,56 +154,6 @@ class Dataset(NDCube):
         Convenience attribute to acces the inventory metadata.
         """
         return self.meta['inventory']
-
-    """
-    Dataset loading and saving routines.
-    """
-
-    @classmethod
-    def from_directory(cls, directory):
-        """
-        Construct a `~dkist.dataset.Dataset` from a directory containing one
-        asdf file and a collection of FITS files.
-        """
-        base_path = Path(directory).expanduser()
-        if not base_path.is_dir():
-            raise ValueError("directory argument must be a directory")
-        asdf_files = tuple(base_path.glob("*.asdf"))
-
-        if not asdf_files:
-            raise ValueError("No asdf file found in directory.")
-        elif len(asdf_files) > 1:
-            raise NotImplementedError("Multiple asdf files found in this "
-                                      "directory. Use from_asdf to specify which "
-                                      "one to use.")  # pragma: no cover
-
-        asdf_file = asdf_files[0]
-
-        return cls.from_asdf(asdf_file)
-
-    @classmethod
-    def from_asdf(cls, filepath):
-        """
-        Construct a dataset object from a filepath of a suitable asdf file.
-        """
-        from dkist.dataset import TiledDataset
-        filepath = Path(filepath).expanduser()
-        base_path = filepath.parent
-        try:
-            with importlib_resources.as_file(importlib_resources.files("dkist.io") / "level_1_dataset_schema.yaml") as schema_path:
-                with asdf.open(filepath, custom_schema=schema_path.as_posix(),
-                               lazy_load=False, copy_arrays=True) as ff:
-                    ds = ff.tree['dataset']
-                    if isinstance(ds, TiledDataset):
-                        for sub in ds.flat:
-                            sub.files.basepath = base_path
-                    else:
-                        ds.files.basepath = base_path
-                    return ds
-
-        except ValidationError as e:
-            err = f"This file is not a valid DKIST Level 1 asdf file, it fails validation with: {e.message}."
-            raise TypeError(err) from e
 
     """
     Private methods.
--- a/dkist/dataset/dataset.py
+++ b/dkist/dataset/dataset.py
@@ -138,16 +138,21 @@ class Dataset(NDCube):
         idx = self.files._array_slice_to_loader_slice(slice_)
         if idx == (np.s_[:],):
             return self.headers.copy()
-        file_idx = []
-        for i in idx:
-            if not isinstance(i, slice):
-                i = slice(i, i+1)
-            file_idx.append(i)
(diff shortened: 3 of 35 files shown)


SolarDrew:
This doesn't actually handle loading multiple datasets yet as requested in #128 but it does unify the API. How important do we think multi-dataset loading is? Because unless we think it's high-priority then this can be merged when it's approved and the rest can be added later without breaking the user experience.

SolarDrew:
@Cadair any objections to me merging this now it's passing?

dkist/dataset/dataset.py
```diff
@@ -192,56 +188,6 @@ def inventory(self):
         """
         return self.meta['inventory']
 
-    """
-    Dataset loading and saving routines.
-    """
-
-    @classmethod
-    def from_directory(cls, directory):
```
Cadair, line 200: I think we should probably deprecate these for a while. I know we are technically in beta but we showed them at the workshop and stuff. The implementation of the functions can just call out to the new function and stick an astropy deprecated decorator on top.

dkist/dataset/loader.py
```diff
@@ -0,0 +1,73 @@
...
+from functools import singledispatch
+
+from jsonschema.exceptions import ValidationError
+
+import asdf
+
+
+@singledispatch
+def load_dataset():
+    """Function to load a Dataset from either an asdf file path or a directory."""
```
Cadair, line 12: This should raise an unknown type error as I think it gets called for any unknown types.

dkist/dataset/loader.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+@load_dataset.register
+def _load_from_string(path: str):
+    return _load_from_path(Path(path))
```
Cadair, line 17: we will need to consider how we handle URLs here at somepoint.

dkist/dataset/loader.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+@load_dataset.register
+def _load_from_path(path: Path):
+    path = path.expanduser()
+    if not path.is_dir():
+        if not path.exists():
+            raise ValueError(f"File {path} does not exist.")
+        # elif raise error if extension is not .asdf
```
Cadair, line 26: better to just try it and see I think.
```suggestion
```

dkist/dataset/__init__.py
```diff
@@ -1,2 +1,3 @@
 from .dataset import Dataset
+from .loader import load_dataset
```
Cadair, line 2: can we bubble this up to `dkist.load_dataset` as well. Also are there not examples in the docs which will need updating?

dkist/dataset/loader.py
```diff
@@ -0,0 +1,73 @@
+import importlib.resources as importlib_resources
+from pathlib import Path
+from functools import singledispatch
+
+from jsonschema.exceptions import ValidationError
```
Cadair, line 5: we might as well take this opportunity to fix #276 and add jsonschema to the setup.cfg file.

dkist/dataset/loader.py
```diff
@@ -0,0 +1,73 @@
...
+
```
Cadair, line 6:
```suggestion
```

Cadair (changes requested):

dkist/dataset/__init__.py
SolarDrew: Thought I'd updated the docs, because the builds were failing. But I'll check again.

changelog/274.feature.rst
```diff
@@ -0,0 +1 @@
+Add a new `load_dataset()` function to combine and replace `Dataset.from_directory()` and `Dataset.from_asdf()`.
```
Cadair, line 1:
```suggestion
Add a new `dkist.load_dataset` function to combine and replace ``Dataset.from_directory()`` and ``Dataset.from_asdf()``.
```

dkist/dataset/tests/test_dataset.py
```diff
@@ -71,28 +72,49 @@ def test_dimensions(dataset, dataset_3d):
...
+        load_dataset(rootdir / "notadirectory")
         assert "directory argument" in str(e)
 
 
+def test_load_tiled_dataset():
+    ds = load_dataset(os.path.join(rootdir, 'test_tiled_dataset-1.0.0_dataset-1.1.0.asdf'))
+    assert ds.shape == (3, 3)
+
+
+def test_load_with_old_methods():
```
SolarDrew, line 99: These tests appear to be running and passing, but for some reason codecov claims that the contents of `from_directory` and `from_asdf` aren't getting hit. Not sure what that's about.

dkist/dataset/loader.py
```diff
@@ -0,0 +1,75 @@
...
+import asdf
+
+
+@singledispatch
+def load_dataset(target):
+    """Function to load a Dataset from either an asdf file path or a directory."""
+    raise TypeError("""Input type not recognised. Must be a string or pathlib.Path referencing a
+        .asdf file or a directory containing one.""")
```
Cadair, line 14: This indentation makes my head hurt.
```suggestion
    raise TypeError("Input type not recognised. Must be a string or pathlib.Path referencing a
                    .asdf file or a directory containing one.")
```

dkist/dataset/tests/test_dataset.py
```diff
@@ -71,38 +72,65 @@ def test_dimensions(dataset, dataset_3d):
...
 
 def test_from_not_directory():
     with pytest.raises(ValueError) as e:
-        Dataset.from_directory(rootdir / "notadirectory")
+        load_dataset(rootdir / "notadirectory")
         assert "directory argument" in str(e)
 
 
+def test_load_tiled_dataset():
+    ds = load_dataset(os.path.join(rootdir, 'test_tiled_dataset-1.0.0_dataset-1.1.0.asdf'))
```
Cadair, line 95:
```suggestion
    ds = load_dataset(os.path.join(rootdir, 'test_tiled_dataset-1.0.0_dataset-1.1.0.asdf'))
    assert isinstance(ds, TiledDataset)
```

Cadair (approved):
Couple of super minor comments, but then merge away.

dkist/dataset/tests/test_dataset.py
```diff
@@ -12,13 +12,14 @@
 from astropy.tests.helper import assert_quantity_allclose
 
 from dkist.data.test import rootdir
-from dkist.dataset import Dataset
+from dkist.dataset import Dataset, load_dataset
```
Cadair, line 15:
```suggestion
from dkist.dataset import Dataset, load_dataset, TiledDataset
```

dkist/dataset/loader.py
```diff
@@ -0,0 +1,75 @@
...
+from jsonschema.exceptions import ValidationError
+
+import asdf
+
+
+@singledispatch
+def load_dataset(target):
+    """Function to load a Dataset from either an asdf file path or a directory."""
+    raise TypeError("Input type not recognised. Must be a string or pathlib.Path referencing a
+                    .asdf file or a directory containing one.")
```
Cadair, line 14:
```suggestion
    raise TypeError("Input type not recognised. Must be a string or pathlib.Path referencing a "
                    ".asdf file or a directory containing one.")
```

Cadair:
@pre-commit.ci autofix

PR #275, opened 2023-07-05
Fix slicing headers with whole axis

Cadair:
@SolarDrew Can you add tests for this asap, because we probably shouldn't ship it if there are more hidden bugs.


dkist/conftest.py
```diff
@@ -290,3 +292,14 @@ def small_visp_dataset():
     vispdir = Path(rootdir) / "small_visp"
     with asdf.open(vispdir / "test_visp.asdf") as f:
         return f.tree['dataset']
+
+
+@pytest.fixture(scope="session")
+def large_visp_dataset():
+    vispdir = Path(rootdir) / "large_visp"
```
Cadair, line 303: think we would be better using a tmppath here.

dkist/conftest.py
```diff
@@ -290,3 +292,14 @@ def small_visp_dataset():
...
+    ds = SimpleVISPDataset(n_maps=1, n_steps=20, n_stokes=4, time_delta=10,
+                           linewave=500*u.nm, detector_shape=(500, 1280))
```
Cadair, line 305: Make the detector shape smaller and it will take less time to save out the files, we don't really need the array shape to be realistic for anything do we?

tox.ini
```diff
@@ -21,6 +21,8 @@ description =
     devdeps: with the latest developer version of key dependencies
     oldestdeps: with the oldest supported version of all dependencies
 deps =
+    git+https://bitbucket.org/dkistdc/dkist-data-simulator
+    git+https://bitbucket.org/dkistdc/dkist-inventory
```
Cadair, line 25: The pypi versions of these should be fine (they are both on pypi).

tox.ini
```diff
@@ -15,12 +15,15 @@ setenv =
 extras = tests
 commands =
     pip freeze --all --no-input
+    python -c 'from astropy.utils.iers import IERS_Auto; IERS_Auto.open()'
```
Cadair, line 18:
```suggestion
    # pre-cache the IERS file for astropy to prevent downloads with remote_data off
    python -c 'from astropy.utils.iers import IERS_Auto; IERS_Auto.open()'
```

tox.ini
```diff
@@ -15,12 +15,15 @@ setenv =
...
     {env:PYTEST_COMMAND} {posargs}
 description =
     run tests
     devdeps: with the latest developer version of key dependencies
     oldestdeps: with the oldest supported version of all dependencies
 deps =
+    dkist-data-simulator
+    dkist-inventory
```
Cadair, line 26: These should be in setup.cfg under the tests extra.

setup.cfg
```diff
@@ -22,18 +22,19 @@ packages = find:
...
-  matplotlib>=3.4
+  matplotlib>=3.5
   ndcube[plotting,reproject]>=2.0
-  numpy>=1.21
-  parfive[ftp]>=1.3
-  sunpy[net,asdf]>=4
+  numpy>=1.22
+  parfive[ftp]>=1.5
+  sunpy[net,asdf]>=4.0.7
+  setuptools>=59 # Minimum version older than 60 to avoid distutils warnings
```
Cadair, line 37: aren't this line and comment counter to each other?

SolarDrew: >=

:shrug: Going from the previous comment, I don't actually know which versions are the ones giving the warnings

Cadair:
```suggestion
  setuptools>=59
```

PR #277, opened 2023-07-20
Varying celestial transforms 3 d

svensk-pojke:
Adds 3D capability to VaryingCelestialTransform and VaryingCelestialTransformSlit. Submitting PR now to facilitate deployment of Cryonirsp pipeline on DKIST DC stacks. A later upgrade to this code will include additional unit tests for transforms to include round trip pixel-to-world --> world to pixel testing. In addition, I will consolidate the many different VaryingCelestialTransform classes into far fewer, and evaluate the possibility of generic N dimensional VaryingCelestialTransform  and VaryingCelestialTransformSlit classes.

### DIFF
--- a/changelog/277.feature.rst
+++ b/changelog/277.feature.rst
@@ -1,0 +1,1 @@
+Add experimental support for 3D LUTs to TimeVaryingCelestialTransform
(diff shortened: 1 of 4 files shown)


Cadair (approved):
As long as you pinky promise to come back and write more tests then it's good with me :wink:.

Can you add changelog fragment saying that "experimental support" or similar has been added for 3D table models?

A-Derks (approved):

PR #282, opened 2023-07-31
move iers pre-fetch to conftest.py from tox.ini

braingram:
asdf includes dkist in it's downstream testing and has noticed recent failures during test collection due to remote data access during tests.
https://github.com/asdf-format/asdf/actions/runs/5716903417/job/15489509194?pr=1597#step:10:684

While looking into the issue I noticed the recent addition tox.ini:
https://github.com/DKISTDC/dkist/blob/25af24a17455aa4094ecb031f4d138cb2732ba5c/tox.ini#L18-L19

As an alternative, this PR moves the fetch to `pytest_configure` in [conftest.py](https://github.com/DKISTDC/dkist/blob/25af24a17455aa4094ecb031f4d138cb2732ba5c/conftest.py) to allow this pre-fetch to occur during both:
- downstream testing in other packages which don't use the tox.ini of this package (for example asdf)
- local runs that occur outside of tox (I was able to replicate the error above locally by calling `astropy.utils.data.clear_download_cache` prior to the test run).

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,3 +1,10 @@
 import matplotlib
+import pytest
 
 matplotlib.use("Agg")
+
+
+def pytest_configure(config):
+    # pre-cache the IERS file for astropy to prevent downloads
+    # which will cause errors with remote_data off
+    from astropy.utils.iers import IERS_Auto; IERS_Auto.open()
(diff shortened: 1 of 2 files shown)


Cadair (approved):
Thanks a lot for the PR!

braingram:
@Cadair I rebased this thinking that was preventing the auto-merge and the push appears to have disabled the auto-merge.

Let me know if there's anything you'd like me to do to move this PR along. Thanks!

PR #294, opened 2023-08-10
Add missing dependencies to setup.cfg

### LINKED ISSUE #291: Missing dependancies
We import `aiohttp` and `tqdm` directly but only rely on them via transitive dependencies.

SolarDrew:
Fixes #291


setup.cfg
```diff
@@ -36,6 +36,8 @@ install_requires =
   sunpy[net,asdf]>=4.0.7
   setuptools>=59
   jsonschema>=3.2
+  aiohttp>=3.6
```
Cadair, line 39: The fact we are doing this for typehints is somewhat annoying but hey.

Cadair:
@SolarDrew The oldestdeps test fail is a deprecation warning from Python, we need to ignore it in the `setup.cfg` file to make the build pass.

SolarDrew:
Would it not be better to remove the deprecated feature and drop 3.9, or are we not ready to do that yet?

Cadair:
The issue is that the oldest version of aiohttp uses the feature deprecated in 3.8

setup.cfg
SolarDrew: We are? I was just going by NEP 29.

PR #295, opened 2023-08-10
import ValidationError from asdf, drop jsonschema as a dependency

braingram:
asdf has recently run into compatibility issues with jsonschema 4.18+. The newer versions of jsonschema dropped support for a feature required in asdf which left us with little choice but to vendorize jsonschema. This presented an opportunity to clean up some leaky bits of the asdf API including the issue that `jsonschema.ValidationError` was transparently passed to user code and libraries (like dkist) had to import `ValidationError` from jsonschema.

asdf 2.15 introduced `ValidationError` as part of the public api available at `asdf.exceptions.ValidationError`. This PR changes the import `ValidationError` from `asdf.exceptions` (instead of jsonschema).

asdf 2.15.1 includes the vendorized jsonschema, keeps jsonschema as a dependency and attempts to use the exceptions from the installed jsonschema (to not break some downstream libraries that use asdf). We are planning to drop jsonschema as a dependency and stop using it's exceptions in asdf 3.0 (or possibly in 2.15.2 if further changes to jsonschema make it's exceptions incompatible).

Please see the [What's New](https://asdf.readthedocs.io/en/latest/asdf/whats_new.html#jsonschema) page in the asdf 2.15.1 docs for more information and let me know if you have any questions, comments or concerns.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,9 +22,9 @@ packages = find:
 include_package_data = True
 install_requires =
   appdirs>=1.4
-  asdf>=2.15.1
+  asdf>=2.9.2
   asdf-astropy>=0.1.1
-  asdf-transform-schemas>=0.3
+  asdf-transform-schemas
   astropy>=5
   dask[array]>=2021.8.0
   globus-sdk>=3.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,7 +22,7 @@ packages = find:
 include_package_data = True
 install_requires =
   appdirs>=1.4
-  asdf>=2.9.2
+  asdf>=2.15.1
   asdf-astropy>=0.1.1
   asdf-transform-schemas
   astropy>=5
@@ -35,7 +35,6 @@ install_requires =
   parfive[ftp]>=1.5
   sunpy[net,asdf]>=4.0.7
   setuptools>=59
-  jsonschema>=3.2
   aiohttp>=3.6
   tqdm>=4.63
 setup_requires = setuptools_scm
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,7 +24,7 @@ install_requires =
   appdirs>=1.4
   asdf>=2.15.1
   asdf-astropy>=0.1.1
-  asdf-transform-schemas
+  asdf-transform-schemas>=0.3
   astropy>=5
   dask[array]>=2021.8.0
   globus-sdk>=3.0
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,11 +1,3 @@
-1.0.1 (unreleased)
-==================
-
-Trivial/Internal Changes
-------------------------
-
-- Import `ValidationError` from `asdf.exceptions` and drop `jsonschema` as a dependency. (`#295 <https://github.com/DKISTDC/dkist/pull/295>`_)
-
 1.0.0 (2023-08-09)
 ==================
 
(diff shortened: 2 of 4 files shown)


setup.cfg
```diff
@@ -22,9 +22,9 @@ packages = find:
 include_package_data = True
 install_requires =
   appdirs>=1.4
-  asdf>=2.9.2
+  asdf>=2.15.1
```
Cadair, line 25: Thanks a lot for the PR. I think this is my main issue with this, I am not sure it's really practical for us to pull in the very latest version of asdf.

Does asdf throw a version warning if you load a file with new asdf or just third party plugins?

braingram: Thanks for taking and look and for the response.

I tested reading and writing asdf files across versions (2.9.2 and 2.15.0). In both cases the file was produced with:
```
t = astropy.time.Time(1.0, format='mjd')
af = asdf.AsdfFile({'t': t})
af.write_to('time.asdf')
```

For 2.9.2 I installed `asdf-astropy==0.1.1` (to match the minimum pin in this package). For 2.15.0 I installed `asdf-astropy==0.4.0`.

Opening the file written with asdf 2.9.2 in asdf 2.15.0 produced no warnings as all the extensions used to produce the file were installed and were at least the same (or in this case newer) versions.

Opening the file written with asdf 2.15.0 in asdf 2.9.2 produced two warnings, one for the asdf-astropy extension and one for the `builtin` extension that handles the asdf core objects:
```
/Users/bgraham/.pyenv/versions/3.10.6/envs/dkist_asdf_downstream/lib/python3.10/site-packages/asdf/asdf.py:348: AsdfWarning: File 'file:///Users/bgraham/Desktop/time.asdf' was created with extension class 'asdf.extension.BuiltinExtension' (from package asdf==2.15.0), but older package (asdf==2.9.2) is installed.
  warnings.warn(msg, AsdfWarning)
/Users/bgraham/.pyenv/versions/3.10.6/envs/dkist_asdf_downstream/lib/python3.10/site-packages/asdf/asdf.py:348: AsdfWarning: File 'file:///Users/bgraham/Desktop/time.asdf' was created with extension URI 'asdf://asdf-format.org/core/extensions/core-1.5.0' (from package asdf-astropy==0.4.0), but older package (asdf-astropy==0.1.1) is installed.
  warnings.warn(msg, AsdfWarning)
```

I'm trying to think of a way to not pin asdf to at least 2.15.0. I can think of three options:

1) Drop jsonschema as a dependency and import `ValidationError` from the top level `asdf`. It appears that this was available at least as [far back as 2.9](https://github.com/asdf-format/asdf/blob/2.9.x/asdf/__init__.py). I should note that having `ValidationError` exposed at the top level `asdf` will likely be deprecated in asdf 3.x.x and removed in 4.0.0. It's probably ambitious to say 4.0.0 will come out in the next 6 months (and as we don't yet have a concrete roadmap for the included changes it will likely take longer) but the deprecation of `asdf.ValidationError` in favor of `asdf.exceptions.ValidationError` will come much sooner and likely soon after the 3.0.0 release (which I'm hoping will happen in the next few months at the latest). This deprecation warning can always be ignored so this does seem like an ok option for keeping the lower pin at 2.9.2 and dropping jsonschema.

2) Keep jsonschema as a dependency and change the `ValidationError` imports to first try from `asdf.exceptions` and then fall back to jsonschema.

3) Drop jsonschema and change the imports to first try from `asdf.exceptions` and then from `asdf`. I kind of like this option the best but I'm happy to make whatever changes you'd like.

I will update this PR for option 3 above but please let me know if you'd rather go with one of the other options or if you have other plans for keeping the pin. Thanks!

Cadair (approved):

PR #305, opened 2023-10-16
Coordinate fixes

Cadair:
This is the dkist part of a multi-project improvement to the coordinate info saved into ASDF.

See Also:

* https://bitbucket.org/dkistdc/dkist-inventory/pull-requests/39
* https://bitbucket.org/dkistdc/dkist-data-simulator/pull-requests/42
* https://github.com/spacetelescope/gwcs/pull/457


SolarDrew:
I'm going to go ahead and assume this isn't expected behaviour:

```
In [1]: import dkist

In [2]: ds = dkist.load_dataset("/home/drew/oss-projects/dkist/dkist/data/test/VISP_L1_20221024T185745_BKEWK_updated.asdf")

In [3]: ds
Out[3]: ---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
File ~/mambaforge/envs/dkist-gwcs-fixed/lib/python3.11/site-packages/IPython/core/formatters.py:708, in PlainTextFormatter.__call__(self, obj)
    701 stream = StringIO()
    702 printer = pretty.RepresentationPrinter(stream, self.verbose,
    703     self.max_width, self.newline,
    704     max_seq_length=self.max_seq_length,
    705     singleton_pprinters=self.singleton_printers,
    706     type_pprinters=self.type_printers,
    707     deferred_pprinters=self.deferred_printers)
--> 708 printer.pretty(obj)
    709 printer.flush()
    710 return stream.getvalue()

File ~/mambaforge/envs/dkist-gwcs-fixed/lib/python3.11/site-packages/IPython/lib/pretty.py:410, in RepresentationPrinter.pretty(self, obj)
    407                         return meth(obj, self, cycle)
    408                 if cls is not object \
    409                         and callable(cls.__dict__.get('__repr__')):
--> 410                     return _repr_pprint(obj, self, cycle)
    412     return _default_pprint(obj, self, cycle)
    413 finally:

File ~/mambaforge/envs/dkist-gwcs-fixed/lib/python3.11/site-packages/IPython/lib/pretty.py:778, in _repr_pprint(obj, p, cycle)
    776 """A pprint that just redirects to the normal repr function."""
    777 # Find newlines and replace them with p.break_()
--> 778 output = repr(obj)
    779 lines = output.splitlines()
    780 with p.group():

File ~/oss-projects/dkist/dkist/dataset/dataset.py:229, in Dataset.__repr__(self)
    225 """
    226 Overload the NDData repr because it does not play nice with the dask delayed io.
    227 """
    228 prefix = object.__repr__(self)
--> 229 output = dedent(f"{prefix}\n{self.__str__()}")
    230 return output

File ~/oss-projects/dkist/dkist/dataset/dataset.py:233, in Dataset.__str__(self)
    232 def __str__(self):
--> 233     return dataset_info_str(self)

File ~/oss-projects/dkist/dkist/dataset/utils.py:67, in dataset_info_str(ds)
     64 for iwrl in range(wcs.world_n_dim):
     66     name = wcs.world_axis_names[::-1][iwrl] or 'None'
---> 67     typ = wcs.world_axis_physical_types[::-1][iwrl] or 'None'
     68     unit = wcs.world_axis_units[::-1][iwrl] or 'unknown'
     70     s += (('{0:' + str(world_dim_width) + 'd}').format(iwrl) + '  ' +
     71             ('{0:' + str(world_nam_width) + 's}').format(name) + '  ' +
     72             ('{0:' + str(world_typ_width) + 's}').format(typ) + '  ' +
     73             '{:s}'.format(unit + '\n'))

IndexError: tuple index out of range
```

dkist/io/asdf/converters/dataset.py
```diff
@@ -27,6 +28,21 @@ def from_yaml_tree(self, node, tag, ctx):
         unit = node.get("unit")
         mask = node.get("mask")
 
+        # If we have a tag older than 1.2.0 then we are going to see if we can
```
Cadair, line 32: @braingram Is there a way I can do this by overriding the Tabular1D converter? or is this the "best" available option?

dkist/io/asdf/resources/manifests/dkist-1.2.0.yaml
```diff
@@ -0,0 +1,14 @@
...
+id: asdf://dkist.nso.edu/dkist/manifests/dkist-1.2.0
+extension_uri: asdf://dkist.nso.edu/dkist/extensions/dkist-1.0.0
+title: DKIST extension
+description: ASDF schemas and tags for DKIST classes.
+
+tags:
+  - schema_uri: "asdf://dkist.nso.edu/schemas/file_manager-1.0.0"
+    tag_uri: "asdf://dkist.nso.edu/tags/file_manager-1.0.0"
+  - schema_uri: "asdf://dkist.nso.edu/schemas/dataset-1.2.0"
+    tag_uri: "asdf://dkist.nso.edu/tags/dataset-1.2.0"
```
Cadair, line 12: @braingram When making a new manifest like this, should it have the old tags in it as well or is it better to leave those to the old manifests?

dkist/io/asdf/converters/dataset.py
Cadair: Thanks!

dkist/io/asdf/resources/manifests/dkist-1.2.0.yaml
braingram: In this case it looks like it's better to leave those in the old manifests and not include them in the new one. As long as the old manfiests/extensions are registered (after the new extension) asdf will use whatever old manifest/extension contains the old tag when it reads a file with that tag. On write it will use the first extension that supports the type (which is why the new extension should be registered first). If you were to include the old tags you would need to update `select_tag` in the `Converter` to pick the appropriate tag (either the new or old version). Otherwise including the old tags would result in writing only the old tags.

Cadair: Thanks!

SolarDrew (approved):
LGTM :+1:  I'm still playing a bit with getting the updated asdfs through my crop tests, but they've taken everything I've been able to through at them so far and the remaining issues are at my end so I don't see why they should block this being merged.

dkist/utils/_model_to_graphviz.py
```diff
@@ -0,0 +1,181 @@
...
+
+@to_subgraph.register(Model)
+def model_to_subgraph(model, inputs=None, outputs=None):
+    inputs = inputs or model.inputs
+    outputs = outputs or model.outputs
+
+    input_labels = inputs
+    output_labels = outputs
+
+    if isinstance(model, Mapping):
```
eigenbrot, line 42: This line makes me think we need another dispatch function registered for a `Mapping` argument. (It would just call this `Model` function).

setup.cfg
```diff
@@ -41,8 +41,8 @@ setup_requires = setuptools_scm
 
 [options.extras_require]
 tests =
-  dkist-data-simulator
-  dkist-inventory
+  dkist-data-simulator@git+https://bitbucket.org/dkistdc/dkist-data-simulator@visp_wcs
```
eigenbrot, line 44: Make sure to update these once all PRs are merged

eigenbrot (approved):

dkist/utils/_model_to_graphviz.py
```diff
@@ -0,0 +1,181 @@
+"""
+This module vizualises compound models by rendering them as a graphviz graph.
+
+It is provided as a helper module which is useful for debugging complex compound models.
+
+It is not tested, incomplete and *should be used at your own risk
```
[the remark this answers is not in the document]
Cadair:
```suggestion
It is not tested, incomplete and *should be used at your own risk*.
```

dkist/utils/_model_to_graphviz.py
Cadair: I think that would be hard to do because you need to just change `output_labels` not outputs, so you would have to refactor things more than I am willing to do for a debug script :wink:

changelog/305.bugfix.1.rst
```diff
@@ -0,0 +1 @@
+Fix the oversight that when generating a model for a celestial WCS the scale model was put before the affine transform in the pipeline. This means that the units for the affine transform matrix provided to ``VaryingCelestialTransform`` and ``generate_celestial_transform`` should be pixels not degrees.
```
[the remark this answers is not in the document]
Cadair: I am not sure I see it, can you add a suggestion with exactly what you mean?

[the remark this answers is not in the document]
eigenbrot: What I was thinking was "Fix the oversight where, when generating a model for a celestial WCS, the scale model was put before the affine transform in the pipeline."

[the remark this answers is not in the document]
Cadair:
```suggestion
Fix the oversight where, when generating a model for a celestial WCS, the scale model was put before the affine transform in the pipeline. This means that the units for the affine transform matrix provided to ``VaryingCelestialTransform`` and ``generate_celestial_transform`` should be pixels not degrees.
```

setup.cfg
Cadair: I think these might have to stay until those branches are released. This is a real sticky circular dependency issue.

Cadair, line 44:
```suggestion
  dkist-data-simulator>=4.1.0rc1
```

PR #308, opened 2023-10-20
add asdf-astropy to devdeps, update schemas to use tag wildcards, add warning ignore for gwcs 0.18.3

braingram:
The devdeps job is failing due to testing using astropy dev (which contains some changes to where classes are implemented, specifically `CartesianRepresentation`):
https://github.com/DKISTDC/dkist/actions/runs/6509684073/job/17681742351#step:10:734
This is resulting in a `ValidationError` as asdf fails to serialize the moved `CartesianRepresentation` as the released version of asdf-astropy is not aware of the new class location.

This PR adds the development version of asdf-astropy to the devdeps job as it contains updates for compatibility with the development version of astropy including: https://github.com/astropy/asdf-astropy/pull/181

During testing it was revealed that several dkist schemas have exact tag requirements:
https://github.com/DKISTDC/dkist/blob/b95f22a5f7ea320d761e7a4834755c2077ea7c26/dkist/io/asdf/resources/schemas/dataset-1.1.0.yaml#L34
this was causing test failures as asdf-astropy dev contains a new `table-1.1.0`. This PR was updated to use wildcards for these tag validation checks (using `table-1.*` in this case).

Note that wildcards were only added to non-dkist tag references and not to dkist tags like:
https://github.com/DKISTDC/dkist/blob/b95f22a5f7ea320d761e7a4834755c2077ea7c26/dkist/io/asdf/resources/schemas/dataset-1.1.0.yaml#L14

Additionally, as the CI was failing due to a warning from gwcs 0.18.3 a warning filter was added. This should be removed when the upper pin for gwcs can be removed.

### DIFF
--- a/changelog/308.trivial.rst
+++ b/changelog/308.trivial.rst
@@ -1,0 +1,3 @@
+To improve compatibility with external libraries that provide ASDF serialization and
+validation (like asdf-astropy) dkist schemas were updated to use tag wildcards
+when checking tagged objects (instead of requiring specific tag versions).
--- a/dkist/io/asdf/resources/schemas/dataset-1.0.0.yaml
+++ b/dkist/io/asdf/resources/schemas/dataset-1.0.0.yaml
@@ -15,17 +15,15 @@ properties:
 
   wcs:
     description: The coordinate system for the complete dataset.
-    anyOf:
-      - tag: "tag:stsci.edu:gwcs/wcs-1.0.0"
-      - tag: "tag:stsci.edu:gwcs/wcs-1.1.0"
+    tag: "tag:stsci.edu:gwcs/wcs-1.*"
 
   mask:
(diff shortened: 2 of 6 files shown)


braingram:
I am now seeing failures in the devdeps due to dkist schema expecting a particular tag version for tables:
https://github.com/search?q=repo%3ADKISTDC%2Fdkist%20tag%3Aastropy.org%3Aastropy%2Ftable%2Ftable-1.0.0&type=code
The development version of asdf-astropy contains a new table schema version [1.1.0](https://github.com/astropy/asdf-astropy/blob/main/asdf_astropy/resources/schemas/table/table-1.1.0.yaml) which is being used and causing `ValidationError`s.
Perhaps the dkist schema can use tag wildcards instead of exact versions?

Cadair (approved):
Thanks a lot @braingram

PR #310, opened 2023-11-08
Tests for Dataset.crop()

SolarDrew:
Adds test for various different use-cases of crop with a couple of datasets from different instruments. Depends on spacetelescope/gwcs#457 for a number of gWCS fixes.


tox.ini
```diff
@@ -61,6 +61,11 @@ commands_pre =
     oldestdeps: pip install -r requirements-min.txt cryptography<42 jsonschema==4.0.1
     figure: python -c "from dkist.data.sample import download_all_sample_data; download_all_sample_data()"
     pip freeze --all --no-input
+
+[testenv:build_docs]
+changedir = docs
+extras = docs
+description = Invoke sphinx-build to build the HTML docs
```
Cadair, line 68:
```suggestion
```

Cadair:
This has suspiciously worked?

dkist/dataset/tests/test_crop_dataset.py
```diff
@@ -0,0 +1,180 @@
...
+              croppable_cryo_dataset.wcs.pixel_to_world(200, 200, 2, 2, 3))
+    coord0 = SkyCoord(-1011.1*u.arcsec, 314.1*u.arcsec,
+                      frame=coords[0][0].frame)
+    coord1 = SkyCoord(-1039.5*u.arcsec, 297.6*u.arcsec,
+                      frame=coords[1][0].frame)
+
+    # Crop using user-defined coords
+    cropped = croppable_cryo_dataset.crop([
+        coord0,
+        Time("2023-01-01T13:00:04"),
```
Cadair, line 161: Ah the curse of the magic time

dkist/dataset/tests/test_crop_dataset.py
```diff
@@ -0,0 +1,180 @@
...
+    assert (cropped_coords[0][-1] == orig_coords[0][600][500:1001]).all()
+    assert (cropped_coords[1] == orig_coords[1]).all()
+    assert (cropped_coords[2] == orig_coords[2][200:601]).all()
+    assert (cropped_coords[3] == orig_coords[3]).all()
+
+
+def test_crop_cryo_by_only_stokes(croppable_cryo_dataset):
+    cropped = croppable_cryo_dataset.crop([
+        None,
+        None,
```
Cadair, line 98: Do you need the `None`?

SolarDrew: Yes, unless I've misunderstood you:

```
(Pdb) croppable_cryo_dataset.crop([StokesCoord('I')], [StokesCoord('I')])
*** ValueError: 1 components in point 0 do not match WCS with 3 components.
```

SolarDrew:
I think this is now just waiting on an actual release of gwcs, assuming the next one is 0.22.2 and includes spacetelescope/gwcs#457 . At that point the devdeps build should pass without any further changes here and everything will be wonderful and lovely.

SolarDrew:
Oh wait that build is passing and it shouldn't be. Never mind.

dkist/dataset/tests/test_crop_dataset.py
```diff
@@ -1,9 +1,14 @@
+import pytest
+
 import astropy.units as u
 from astropy.coordinates import SkyCoord, SpectralCoord, StokesCoord
 from astropy.time import Time
 
+gwcs = pytest.importorskip("gwcs", "0.22.2")
```
Cadair, line 7: I was thinking more in the test which needed this, but if they all do this is fine.

dkist/dataset/tests/test_crop_dataset.py
```diff
@@ -0,0 +1,185 @@
+import pytest
+
+import astropy.units as u
+from astropy.coordinates import SkyCoord, SpectralCoord, StokesCoord
+from astropy.time import Time
+
+gwcs = pytest.importorskip("gwcs", "0.22.2")
```
Cadair, line 7: This was the first version to have the fix:
```suggestion
gwcs = pytest.importorskip("gwcs", "0.22.2a2")
```

Cadair (approved):

PR #311, opened 2023-11-16
Automate registering search parameter descriptions

### LINKED ISSUE #167: Support querying the dataset API for possible values
The `/datasets` API has a way to get [possible values](https://api.dkistdc.nso.edu/datasets/docs#/default/get_search_values_v1_searchValues_get) we should use it with some caching for the search attrs.

See also: DCS-998

SolarDrew:
Fixes #167 

Example output:

```
In [1]: from sunpy.net import Fido, attrs as a; import dkist.net

In [2]: a.dkist.SpatialSampling
Out[2]: 
<class 'dkist.net.attrs.SpatialSampling'>
dkist.net.attrs.SpatialSampling

The average spatial sampling of a dataset.

        Attribute Name        Client           Full Name                                       Description                            
----------------------------- ------ ----------------------------- -------------------------------------------------------------------
averagedatasetspatialsampling DKIST  averageDatasetSpatialSampling The min/max allowable dataset spatial sampling. 0.0-12388.04306084.

In [3]: a.Time
Out[3]: 
<class 'sunpy.net.attrs.Time'>
sunpy.net.attrs.Time

Specify the time range of the query.

Attribute Name    Client   Full Name                            Description                           
-------------- ----------- --------- -----------------------------------------------------------------
all            NOAAIndices all       All values of this type are supported.                           
all            NOAAPredict all       All values of this type are supported.                           
time           DKIST       time      Min: 2022-02-23T19:05:32.338002; max: 2023-11-01T20:51:20.287000.

In [4]: a.dkist.TargetType
Out[4]: 
<class 'dkist.net.attrs.TargetType'>
dkist.net.attrs.TargetType

Name for the object observed by a dataset.

Attribute Name Client Full Name                         Description                         
-------------- ------ --------- ------------------------------------------------------------
quietsun       DKIST  quietsun  A list of target types that can be present within a dataset.
sunspot        DKIST  sunspot   A list of target types that can be present within a dataset.
unknown        DKIST  unknown   A list of target types that can be present within a dataset.
```

At the moment where multiple categorical options are available for a parameter the descriptions are all the same because it's just the description of that parameter taken from [here](https://api.dkistdc.nso.edu/datasets/docs#/default/get_datasets_v1_get). Hopefully at some point that will change - ideally the [search API](https://api.dkistdc.nso.edu/datasets/docs#/default/get_search_values_v1_searchValues_get) would include those descriptions and they can then be directly registered with the client here.


Cadair:
Having taken this over, here are some thoughts to myself:

* I think we need to have a version of the JSON in the repo, which we update before releases so that it doesn't completely fail if the API calls break.
* We should make an API call and save it to the user data dir and then use that to determine if we need to refresh.
* We should have a timeout / error catching on the refresh that should be triggered by the method which registers the values not at import time of the helper module.

Cadair:
This needs unit tests, but is otherwise complete if people want to look over it.

dkist/net/attrs_values.py
```diff
@@ -0,0 +1,173 @@
...
+        search_values = json.load(f)
+
+    search_values = {param["parameterName"]: param["values"] for param in search_values["parameterValues"]}
+
+    attr_values = {}
+    for key, attr in INVENTORY_ATTR_MAP["categorical"].items():
+        attr_values[attr] = [(name, "") for name in search_values[key]["categoricalValues"]]
+
+    for key, attr in INVENTORY_ATTR_MAP["range"].items():
+        attr_values[attr] = [(key, f"Value between {search_values[key+'Min']['minValue']:.5f} and {search_values[key+'Max']['maxValue']:.5f}")]
```
Cadair, line 171:
```suggestion
        attr_values[attr] = [("all", f"Value between {search_values[key+'Min']['minValue']:.5f} and {search_values[key+'Max']['maxValue']:.5f}")]
```

Cadair (approved):

SolarDrew:
LGTM pending test approval :+1:

dkist/net/attrs_values.py
```diff
@@ -0,0 +1,176 @@
...
+
+    attr_values = {}
+    for key, attr in INVENTORY_ATTR_MAP["categorical"].items():
+        attr_values[attr] = [(name, "") for name in search_values[key]["categoricalValues"]]
+
+    for key, attr in INVENTORY_ATTR_MAP["range"].items():
+        attr_values[attr] = [("all", f"Value between {search_values[key+'Min']['minValue']:.5f} and {search_values[key+'Max']['maxValue']:.5f}")]
+
+    # Time - Time attr allows times in the full range but start and end time are given separately by the DKIST API
+    attr_values[sattrs.Time] = [("time", f"Min: {search_values['startTimeMin']['minValue']}; max: {search_values['endTimeMax']['maxValue']}.")]
```
Cadair, line 174:
```suggestion
    attr_values[sattrs.Time] = [("time", f"Min: {search_values['startTimeMin']['minValue']} - Max: {search_values['endTimeMax']['maxValue']}.")]
```

PR #324, opened 2024-02-02
updates for upcoming ASDF standard 1.6.0

braingram:
asdf is working on making the current development version of the ASDF standard (1.6.0) stable (see https://github.com/asdf-format/asdf/pull/1744 for details). As the 1.6.0 version contains new versions of a few core schemas (including [ndarray](https://github.com/asdf-format/asdf-standard/blob/main/resources/schemas/stsci.edu/asdf/core/ndarray-1.1.0.yaml)) many dependent schemas will require updates to be compatible with the new version (if they contain `$ref` or `tag` links to core schemas).

This PR updates the `dkist` schemas to be compatible with ASDF standard 1.6.0. One example of a required update (for ASDF 1.6.0 compatibility) is updating the `transform` `$ref`s found in several `dkist` schemas to [transform-1.3.0](https://github.com/asdf-format/asdf-transform-schemas/blob/main/resources/stsci.edu/schemas/transform-1.3.0.yaml). The schema updates were accompanied by creation of a new `dkist-wcs-1.3.0` extension.

Similar updates will be needed for `sunpy` (see: https://github.com/sunpy/sunpy/pull/7432) to allow `dkist` to be fully compatible with ASDF standard 1.6.0. (see this run where asdf switched the default to 1.6.0 and dkist tests were run with this PR and the sunpy PR: https://github.com/asdf-format/asdf/actions/runs/8345268069/job/22843626280?pr=1744).

A few minor schema fixes were also added:
- fix manifest `id`s to match `uri`s used to register manifests with asdf
- use a unique `extension_uri` for each extension
- fix tag patterns in `varying_celestial_transform` schema

Several dependency changes are also included (more on this below):
- remove `asdf-unit-schemas` as a dependency (this package only provided redundant schemas and has been decommissioned: https://github.com/asdf-format/asdf-unit-schemas)
- update the minimum required `asdf` to 2.14.4. This was needed for 2 reasons:
  - the `pytest-asdf` plugin to test the examples in the new schemas (which use `asdf-standard` requirements)
  - old versions of `asdf` included experimental support for the development standard (1.6.0). However versions before 2.14 fail to fully support the current state of the ASDF 1.6.0 standard (they will write out files using the old `ndarray-1.0.0` tag instead of the required `ndarray-1.1.0` tag). If a user forced asdf to use the development standard they will have produced files that may not be compatible with the current ASDF 1.6.0 standard. This also means that if the `minimum-dependencies` CI attempts to use a version of asdf < 2.14 that it will produce invalid files and errors.
- increase the minimum required `asdf-transform-schemas` version to allow the `dkist` schema tests in the `minimum-dependencies` job to find the new `transform-1.3.0` schema (also increase the `asdf-standard` version to match the minimum in that version of `asdf-transform-schemas`)
- increase the minimum required `asdf-astropy` version to make it aware of the updated `asdf-transform-schemas` to allow it to test the new schemas during the `minimum-dependencies` job (and increase the `asdf-coordinate-schemas` version to match the requirement in `asdf-astropy`).

Many of the minimum version bumps above were required to allow the schema tests to pass in the `minimum-dependencies` job. If these version increases are not an issue I think they are overall improvements (and reduce the risk that say an old environment is updated and does not pick up the required schemas). If the minimum version increases are an issue one thing to consider would be disabling the schema tests during the `minimum-dependencies` job.


Cadair:
Hi @braingram Thanks so much for this PR (and the sunpy one). The code changes look good, but I have a couple of questions mainly so I understand this properly.

I generally try and stick to SPEC-0 dependency rules (so 2 years for most deps), but I am willing to not do this for ASDF things where it makes sense, because that seems to be a challenge. 

With regards to the minimum dependencies, if we have those new minimums installed would we write files which use 1.6.0? or would it be the minimum needed to read a file written with 1.6.0?

Cadair:
> To make sure I'm understanding, you'd like the minimum versions to be the latest version as of 2 years ago?

That's the default position. As I say though I am happy for there to be well reasoned exceptions.

> This might be possible but I suspect that this PR will instead need to restrict dkist to using ASDF standard 1.5.0 as asdf 2.11 produces invalid files if told to write out 1.6.0.

Under what conditions would asdf 2.11 write out a 1.6.0 file?

> Currently 1.6.0 is not yet the default in asdf (switching that will occur at the next major version change, 4.0, which we are hoping to release in the next 6 months but don't have a hard ETA). So a call to [AsdfFile.write_to](https://asdf.readthedocs.io/en/latest/api/asdf.AsdfFile.html#asdf.AsdfFile.write_to) with no `version` argument will write a 1.5.0 file.

So asdf wont write a 1.6.0 file by default until we write a file out with asdf 4.0+ installed?

Cadair:
Oh the other thing I should make clear is that we write out files with the minimum versions of all these packages at the data center, so that we don't throw warnings about the users env being too old when it is within the minimums for the given version of the `dkist` package.

So I am also trying to understand what the impact would be on us starting to generate our production files with these new minimums.

tox.ini
```diff
@@ -60,8 +60,12 @@ commands =
...
     # Then specify a specific one like this
     # online: {env:PYTEST_COMMAND} --remote-data=any {posargs}
-    !online: {env:PYTEST_COMMAND} {posargs}
-    online: {env:PYTEST_COMMAND} --remote-data=any {posargs}
+    {env:PYTEST_COMMAND} \
+    online: --remote-data=any \
+    # It's not possible to test the new schemas with the oldest dependencies
+    # as the new schemas require new dependent schemas
+    oldestdeps: -o asdf_schema_tests_enabled=false
+    {posargs}
```
Cadair, line 68: oh hey, I didn't know you could do that :thinking:

Cadair:
Thanks a lot for this @braingram and taking the time to explain it all to me.

Cadair (approved):

PR #344, opened 2024-03-08
Refactor VaryingCelestialTransform models

SolarDrew:
Starts compressing WCS models into a single model that can handle multiple dimensions of input.


dkist/wcs/models.py
```diff
@@ -155,6 +157,14 @@ def __init__(self, *args, crval_table=None, pc_table=None, projection=m.Pix2Sky_
             raise TypeError("The projection keyword should be a Pix2SkyProjection model class.")
         self.projection = projection
 
+        self.inputs = ["x", "y", "z", "q", "m"][:self.n_inputs]
+        self.outputs = ("lon", "lat")
```
Cadair, line 161:
```suggestion
        self.outputs = ["lon", "lat"]
```

dkist/wcs/models.py
```diff
@@ -155,6 +157,14 @@ def __init__(self, *args, crval_table=None, pc_table=None, projection=m.Pix2Sky_
             raise TypeError("The projection keyword should be a Pix2SkyProjection model class.")
         self.projection = projection
 
+        self.inputs = ["x", "y", "z", "q", "m"][:self.n_inputs]
+        self.outputs = ["lon", "lat"]
```
Cadair, line 161:
```suggestion
        self.inputs = ("x", "y", "z", "q", "m")[:self.n_inputs]
        self.outputs = ("lon", "lat")
```

dkist/wcs/models.py
```diff
@@ -143,18 +143,31 @@ def sanitize_index(ind):
...
-            self.crval_table
+            self.crval_table,
         ) = self._validate_table_shapes(np.asanyarray(pc_table), np.asanyarray(crval_table))
+        self.is_inverse = inverse
 
         if not isinstance(projection, m.Pix2SkyProjection):
             raise TypeError("The projection keyword should be a Pix2SkyProjection model class.")
         self.projection = projection
 
+        if self.is_inverse:
```
Cadair, line 159: I am conflicted about this. I like the reduction in classes, but I think that ultimately this might be more confusing when introspecting the models etc. The most common pattern with astropy models is that you have a different class for the inverse.

What would the fall out be from putting the inverse classes back as well?

SolarDrew: Minimal I should think. The way it was before there was some calulation happening in the inverse classes but now that's been brought in to the base class it would be pretty straightforward to just have an inverse class for each dimension that sets `self.is_inverse = True` and does nothing else. Some of the corresponding infrastructure would have to go back in but that's fine.

dkist/wcs/models.py
```diff
@@ -14,14 +16,10 @@
 
 __all__ = [
     "CoupledCompoundModel",
-    "InverseVaryingCelestialTransform",
-    "InverseVaryingCelestialTransform2D",
-    "InverseVaryingCelestialTransform3D",
```
Cadair, line 19: revert?

Cadair (approved):

PR #346, opened 2024-03-11
bump gwcs version check in test

braingram:
gwcs 0.21.0 was released.

I believe this may be interfering with a version check in the `test_read_wcs_with_backwards_affine` test.

The check was previously updated in:
https://github.com/DKISTDC/dkist/pull/313
and added in:
https://github.com/DKISTDC/dkist/pull/305
which mentions the draft gwcs PR:
https://github.com/spacetelescope/gwcs/pull/457
which might be what this test is attempting to check against. @Cadair can you confirm?


Cadair:
Thanks @braingram :pray:

PR #354, opened 2024-03-26
Move sunpy imports so test collection works

Cadair:
This combined with https://github.com/sunpy/sunpy/pull/7527 should fix our devdeps build


Cadair:
Alas no.

PR #355, opened 2024-03-26
Remove pyargs because it currently interacts poorly with pytest-asdf

Cadair:
Also @SolarDrew hates it, and we don't have compiled extensions so the best argument for it is void.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -13,9 +13,6 @@ envlist =
 
 [testenv]
 pypi_filter = https://raw.githubusercontent.com/sunpy/sunpy/main/.test_package_pins.txt
-# Run the tests in a temporary directory to make sure that we don't import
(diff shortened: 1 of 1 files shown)


braingram:
FWIW this package doesn't have the same issue as sunpy.

The schema tests are running:
https://github.com/DKISTDC/dkist/actions/runs/8435154989/job/23099969053#step:10:200
as they are found by pytest at the location defined by `asdf_schema_root` (and unlike sunpy this doesn't change for each tox environment).

Cadair:
> and unlike sunpy this doesn't change for each tox environment

eh? Why does sunpy's do that?

braingram:
:shrug: Not sure. I suspect it's to make sure the tests run are not from the source directory.

PR #349, opened 2024-03-19
Update tutorial to use new DDT data and download FITS files in the RTD build

Cadair:
Fixes #348

PR #360, opened 2024-04-02
[WIP] Dataset bounding box example

Cadair:
This PR is based on top of #349.


docs/howto_guides/bounding_boxes_from_dataset.md
```diff
@@ -0,0 +1,93 @@
...
+
+Extract the bottom left and top right corners of the VISP data by finding the world coordinates of the bottom left pixel and top right pixel.
+```{note}
+Note that we subtract 0.5 from the data shape, this is because we want to find the coordinate of the center of the pixel not the edge.
+```
+
+```{code-cell} ipython3
+corners = visp_spatial.wcs.array_index_to_world([0, visp_spatial.data.shape[0] - 0.5],
+                                                [0, visp_spatial.data.shape[1] - 0.5])
+corners
```
Cadair, line 76: I think the fact that `[-0.5, visp_spatial.data.shape[0] - 0.5]` errors here (because of an out of bounds on the time lookup table) is indicative of a bug in the WCS in the ASDFs :sob:

PR #361, opened 2024-04-02
Fix Dask tokenisation performance issues by not hashing

Cadair:
It seems that due to https://github.com/dask/dask/pull/10883 which was included in dask 2024.2.1 there is now a *massive* performance overhead loading large raster ASDF files. All this time seems to be in the tokenisation step, presumably because it tries to actually compute the array or something even if the file is missing.

Based on my reading of [the docs](https://docs.dask.org/en/stable/generated/dask.array.from_array.html#dask.array.from_array) I think we can just get away with setting name to False.

### DIFF
--- a/changelog/361.bugfix.rst
+++ b/changelog/361.bugfix.rst
@@ -1,0 +1,1 @@
+Fix a performance regression when dask>=2024.2.1 is installed.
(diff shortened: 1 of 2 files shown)


SolarDrew (approved):
Seems to do the job :+1:

PR #372, opened 2024-04-17
WIP Start adding some basic benchmarking infrastructure

SolarDrew:
Very rudimentary so far but hopefully will be useful in tracking performance and preventing regressions.

### DIFF
--- a/dkist/tests/test_benchmarks.py
+++ b/dkist/tests/test_benchmarks.py
@@ -1,0 +1,17 @@
+import gzip
+from pathlib import Path
+
+from dkist import load_dataset
+from dkist.data.test import rootdir
+
+
+def load_asdf_from_gzip(tmp_path_factory):
+    vispdir = tmp_path_factory.mktemp("data")
+    with gzip.open(Path(rootdir) / "large_visp.asdf.gz", mode="rb") as gfo:
+        with open(vispdir / "test_visp.asdf", mode="wb") as afo:
+            afo.write(gfo.read())
+    ds = load_dataset(vispdir / "test_visp.asdf")
+
+
+def test_load_asdf(benchmark, tmp_path_factory):
+    benchmark(load_asdf_from_gzip, tmp_path_factory)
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -89,3 +90,8 @@ commands =
     !notebooks: sphinx-build -j 1 --color -W --keep-going -b html -d _build/.doctrees . _build/html -D nb_execution_mode=off {posargs}
     notebooks: sphinx-build -j 1 --color -W --keep-going -b html -d _build/.doctrees . _build/html {posargs}
     python -c 'import pathlib; print("Documentation available under file://\{0\}".format(pathlib.Path(r"{toxinidir}") / "docs" / "_build" / "index.html"))'
+
+[testenv:benchmarks]
+description = Run benchmarks against latest release to ensure there are no performance regressions
+commands =
+    {env:PYTEST_COMMAND} -k test_benchmarks --benchmark-autosave --benchmark-compare=0001
```
Cadair, line 97: I propose that instead of `-k test_benchmarks` we put the [`@pytest.mark.benchmark`](https://pytest-benchmark.readthedocs.io/en/latest/usage.html#markers) marker on all our benchmarks and then do `-m benchmark` to select tests with that marker.

dkist/tests/test_benchmarks.py
```diff
@@ -0,0 +1,17 @@
...
+
+from dkist import load_dataset
+from dkist.data.test import rootdir
+
+
+def load_asdf_from_gzip(tmp_path_factory):
+    vispdir = tmp_path_factory.mktemp("data")
+    with gzip.open(Path(rootdir) / "large_visp.asdf.gz", mode="rb") as gfo:
+        with open(vispdir / "test_visp.asdf", mode="wb") as afo:
+            afo.write(gfo.read())
```
Cadair, line 12: presumeably we would want this bit to be in a fixture so it's not part of the code getting benchmarked?

SolarDrew:
Going to close this as it's been made obsolete by #382 .

PR #382, opened 2024-05-08
Add codspeed workflow to run benchmarks

SolarDrew:
Adds a workflow to run [Codspeed](https://codspeed.io/). Supercedes #372 , as this will hopefully be easier to setup and maintain, and give nicer output, a la codecov.

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -1,0 +1,19 @@
+name: codspeed-benchmarks
+
+on:
+  push:
+    branches:
+      - "main"
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  benchmarks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run benchmarks
+        uses: CodspeedHQ/action@v2
+        with:
+          token: ${{ secrets.CODSPEED_TOKEN }}
+          run: "pytest -vvv -r fEs --pyargs dkist -k test_benchmarks --benchmark-autosave"
--- a/dkist/tests/test_benchmarks.py
+++ b/dkist/tests/test_benchmarks.py
@@ -1,0 +1,17 @@
+import gzip
+from pathlib import Path
+
+from dkist import load_dataset
+from dkist.data.test import rootdir
+
+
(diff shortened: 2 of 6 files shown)


tox.ini
```diff
@@ -10,6 +10,7 @@ envlist =
     py310-oldestdeps
     build_docs{,-notebooks}
     codestyle
+    benchmarks
```
Cadair, line 13:
```suggestion
```

pyproject.toml
```diff
@@ -53,6 +53,7 @@ tests = [
   "pytest-mpl",
   "pytest-httpserver",
   "pytest-filter-subpackage",
+  "pytest-codspeed",
```
Cadair, line 56: As this is really only a CI dependency I would put it in the github actions workflow and not here.

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  benchmarks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
+      - run: python -m pip install .[tests]
```
Cadair, line 18:
```suggestion
      - run: python -m pip install .[tests] pytest-codspeed
```

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,23 @@
...
+  workflow_dispatch:
+
+jobs:
+  benchmarks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
```
Cadair, line 17:
```suggestion
          python-version: '3.12'
```

tox.ini
Cadair, line 13: I think we should add a tox env which uses pytest-benchmark to run the benchmarks and displays some output.

Cadair:
pre-commit.ci autofix

dkist/tests/test_benchmarks.py
```diff
@@ -0,0 +1,27 @@
+
```
Cadair, line 1:
```suggestion
```

PR #383, opened 2024-05-10
Benchmarking tweaks

SolarDrew:
Couple of small changes - add a non-plotting benchmark in the hope that that will prove more useful for coparisons, and only run the plotting ones twice so it doesn't take ages.


SolarDrew:
Well this hasn't made it any quicker it seems :thinking:

PR #370, opened 2024-04-15
More refactoring in VaryingCelestialTransform

SolarDrew:
Slightly re-jig VaryingCelestialTransform to strip units from the inputs in `._map_transform()` and then put them back afterwards. Gives a reasonable performance boost, about 3.5x faster to plot a dataset across non-optimal axes.

PR #387, opened 2024-05-17
Add more benchmarks

SolarDrew:
~~Try slower plot again to see how it does now with performance improvements from #370. I don't expect this to be merged soon, or even worked on necessarily, but I'm interested to see how fast we can get the longer plots running.~~

Turns out the improvements haven't made enough difference to the performance to be able to use the slower plots so this has just become a PR to add some extra benchmarks. Will work on more performance improvements another time and try and use the slower plots when possible.


SolarDrew:
Not sure what else really needs benchmarking, suggestions welcome. Otherwise this is good to go.

dkist/wcs/models.py
```diff
@@ -130,7 +130,7 @@ def update_celestial_transform_parameters(
         -crpix[0],
         -crpix[1],
         pc,
-        transform[2].translation.value,
+        transform[2].translation.value * pc.unit if hasattr(pc, "unit") else transform[2].translation.value,
```
Cadair, line 133: What led to adding this? I think this is the slightly better way to do it though:

```suggestion
        transform[2].translation.quantity if hasattr(pc, "unit") else transform[2].translation.value,
```

Also if this is only getting hit by a benchmark it should have another unit test?

Cadair (approved):
One small tweak otherwise I think this is good.

Cadair (approved):

PR #393, opened 2024-06-13
Fitsio loader

### LINKED ISSUE #159: Consider adding a fitsio backed data loader
I discovered today that Astropy's FITS reader can't handle decompressing individual tiles inside a compressed HDU. This means that when people use the user tools, for any read operation on a file the whole file has to be loaded into memory and decompressed. This is memory and time inefficient. See: https://github.com/astropy/astropy/issues/3895 for more details on the lack of support from Astropy.

An alternative (that I left the door open for) is to use a different FITS library to load the data inside our Dask arrays. The best candidate for this would be the [fitsio](https://github.com/esheldon/fitsio) package, some quick benchmarks indicate that fitsio handles decompressing individual tiles and even decompresses the whole image faster than astropy:

```
In [44]: %timeit fits.getdata("VBI_L1_00656282_2018_05_11T14_25_05_466665_I.fits", hdu=1)
183 ms ± 248 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [45]: %timeit hdu[1][:, :]
131 ms ± 182 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [46]: %timeit f.read("VBI_L1_00656282_2018_05_11T14_25_05_466665_I.fits")
131 ms ± 196 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [48]: %timeit hdu[1][0, 0]
21.6 µs ± 191 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [49]: %timeit hdu[1][0, :]
22.3 µs ± 78.2 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [50]: %timeit hdu[1][:10, :]
235 µs ± 718 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

The advantages of using fitsio include:

* Potentially significant performance enhancements.
* Potentially significant reduction in memory usage.

There are some disadvantages:

* The `fitsio` Python package is GPL v2 licensed, us importing it has potential licencing implications.
* There are no binary wheels of fitsio published on PyPI this means people using PyPI to install this package would have to have a working toolchain and build fitsio.
* The API is lower level, although this wouldn't be exposed to the user.

I think a good approach would be to be able to optionally use fitsio if it is already installed. This way we aren't mandating it's import so we probably could argue the corner with the licencing issues, and also we aren't requiring people to compile it. Making it optional like this would require hooking in some configuration options because the point where the loader is decided is inside the asdf converter, where we can't pass user arguments though.

SolarDrew:
Closes #159 

Adds a FITS loader that uses fitsio. Currently replaces the astropy loader but that's just until I've sorted out the config stuff for switching between them. Mostly making the PR to see how the benchmarks respond because if it's just as slow there's not much point (or the benchmarks don't cover it properly).

### DIFF
--- a/dkist/conftest.py
+++ b/dkist/conftest.py
@@ -22,7 +22,7 @@ from dkist.data.test import rootdir
 from dkist.dataset import Dataset
 from dkist.dataset.tiled_dataset import TiledDataset
 from dkist.io import FileManager
-from dkist.io.loaders import AstropyFITSLoader
+from dkist.io.loaders import FitsioFITSLoader
 
 
 @pytest.fixture
@@ -232,7 +232,7 @@ def dataset(array, identity_gwcs):
     # Construct the filename here as a scalar array to make sure that works as
     # it's what dkist-inventory does
     ds._file_manager = FileManager.from_parts(np.array("test1.fits"), 0, "float", array.shape,
-                                              loader=AstropyFITSLoader)
+                                              loader=FitsioFITSLoader)
 
     return ds
 
--- a/dkist/io/asdf/converters/file_manager.py
+++ b/dkist/io/asdf/converters/file_manager.py
@@ -13,7 +13,7 @@ class FileManagerConverter(Converter):
 
     def from_yaml_tree(self, node, tag, ctx):
         from dkist.io.file_manager import FileManager
-        from dkist.io.loaders import AstropyFITSLoader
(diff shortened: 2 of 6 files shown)


SolarDrew:
Closing this as it seems to make no difference to performance.

PR #394, opened 2024-06-14
Add benchmark for slicing datasets

SolarDrew:


.github/workflows/codspeed.yml
```diff
@@ -16,6 +16,10 @@ jobs:
         with:
           python-version: '3.12'
       - run: python -m pip install .[tests] pytest-codspeed 'numpy<2'
```
Cadair, line 18: We should merge this pr, but without this hack.

Perhaps we should use these two things as "sample data" and integrate them into dkist.data.sample like SunPy?

dkist/tests/test_benchmarks.py
```diff
@@ -35,3 +32,9 @@ def plot_and_save_fig(ds=visp_dataset_no_headers, axes=axes):
         ds.plot(plot_axes=axes)
         plt.savefig("tmpplot")
         plt.close()
+
+
+@pytest.mark.benchmark
+def test_dataset_compute_data(benchmark, real_visp):
```
Cadair, line 38: As an additional benchmark it would be interesting to try one which doesn't load the whole file I.e. slice in either wavelength or slit axes.

dkist/tests/test_benchmarks.py
```diff
@@ -35,3 +32,9 @@ def plot_and_save_fig(ds=visp_dataset_no_headers, axes=axes):
...
+    ds = real_visp[0, :5]
```
Cadair, line 39: Given this benchmark now runs in under 1s maybe this is too little data to get through the overhead of the dask array and stuff? Maybe make it :15 or something?

.github/workflows/codspeed.yml
SolarDrew: I don't really remember how it works in SunPy, but sure. So `dkist.data.sample` would have a function to download these files and we'd call that for running the benchmarks and for the docs build?

dkist/tests/test_benchmarks.py
```diff
@@ -35,3 +32,19 @@ def plot_and_save_fig(ds=visp_dataset_no_headers, axes=axes):
         ds.plot(plot_axes=axes)
         plt.savefig("tmpplot")
         plt.close()
+
+
+@pytest.mark.benchmark
+def test_dataset_compute_data_full_files(benchmark, real_visp):
+    ds = real_visp[0, :15]
+    benchmark(ds.data.compute)
```
Cadair, line 40: Given what we learnt about IO and benchmarking in https://github.com/DKISTDC/dkist/pull/399#issuecomment-2220534572 I think we should add a comment or docstring here noting that this is timing this without the IO overhead.

I think that it's still a useful benchmark, i.e. it would have caught #361 but we should make sure we don't forget what it is actually testing while we are using codspeed.

Cadair:
@SolarDrew you have conflicts. I am working on the sample data thing now.

PR #399, opened 2024-06-21
Fitsioloader plus new benchmark

SolarDrew:
Don't merge this, I'm just opening it to see the codspeed run.


SolarDrew:
Codspeed run says it's not faster, so this can be closed.

braingram:
Does codspeed have a model for io? My understanding is it runs the code through a sort of virtual CPU and looks at what instructions are executed. It looks like it doesn't consider system calls:
https://docs.codspeed.io/features/understanding-the-metrics/#system-calls
but I have been unable to find anything definitive about using it for io heavy operations.

Cadair:
ah yes, it does seem to indicate it would ignore io :frowning:

PR #402, opened 2024-06-24
Assorted improvements for TiledDataset

### LINKED ISSUE #401: Provide an informative repr for TiledDataset

### LINKED ISSUE #403: Expose the shape of all the tiles

### LINKED ISSUE #404: Apply same slice to all tiles in dataset

SolarDrew:
Add `__repr__` (just uses ds[0, 0] for now for ease of putting something together). Closes #401

Add `tiles_shape` property. Closes #403 

Add `slice_tiles()` method to apply the same slice to all datasets. Closes #404


dkist/dataset/tiled_dataset.py
```diff
@@ -124,5 +126,29 @@ def shape(self):
...
         return self._data.shape
 
+    @property
+    def tiles_shape(self):
+        """
+        The shape of each individual tile in the TiledDataset.
+        """
+        return [[tile.data.shape for tile in row] for row in self]
+
+    def slice_tiles(self, slice_):
```
Cadair, line 136: We should have this be a property so you can do `.slice_tiles[10:]` rather than `.slice_tiles(np.s_[10:])`

dkist/dataset/utils.py
```diff
@@ -18,7 +25,11 @@ def dataset_info_str(ds):
     if instr:
```
Cadair, line 25: This has never worked. It should be `ds.inventory.get("instrument")` I think.

Cadair (changes requested):
A couple of things to sort

dkist/dataset/tiled_dataset.py
```diff
@@ -66,6 +81,7 @@ def __init__(self, dataset_array, inventory=None):
         self._data = np.array(dataset_array, dtype=object)
         self._inventory = inventory or {}
         self._validate_component_datasets(self._data, inventory)
+        self._tile_slicer = TiledDatasetSlicer(self._data, self.inventory)
```
Cadair, line 84: I think you should just move this into the property.

dkist/dataset/utils.py
```diff
@@ -10,15 +10,26 @@
...
+    dstype = type(ds).__name__
+    if is_tiled:
+        tile_shape = ds.shape
+        ds = ds[0, 0]
     wcs = ds.wcs.low_level_wcs
 
     # Pixel dimensions table
 
-    instr = ds.meta.get("instrument_name", "")
+    instr = ds.inventory.get("instrument_name", "")
```
Cadair, line 24: If this starting to work doesn't cause a test to fail something has gone wrong.
```suggestion
    instr = ds.inventory.get("instrument", "")
```

Cadair (approved):

PR #407, opened 2024-06-26
Catch URLError in tests when trying to download local attr values

### LINKED ISSUE #381: Fetching local attr values can raise an exception
There is an edge case in the attr values code which can happen (I think) in the following case:

1. Successfully trigger a download of the values to the local data dir.
2. Attempt to refresh the values but have the download fail.
3. Error is raised because the failed download has deleted the pre-existing file.

I think this should be reliably triggered by running:

```bash
$ python -m dkist.net
$ touch -d "2020/01/01T00:00:00" ~/.local/share/dkist/api_search_values.json
$ tox -e py311
```

as the tox refresh will fail due to pytest-remotedata.

SolarDrew:
Closes #381


dkist/net/attrs_values.py
```diff
@@ -105,6 +106,11 @@ def attempt_local_update(*, timeout: int = 1, user_file: Path = None, silence_er
     try:
         _fetch_values_to_file(user_file, timeout=timeout)
         success = True
+    except URLError as urlerr:
```
Cadair, line 109: This is a good addition, but I don't think it solves the general problem as we could still end up in a situation where we later try to access a file we just deleted. We need to rework the logic a bit more I think, to either not delete the file at all (maybe download to a tmp location and then move on success) or to revert to the built in one on fail but that leads to using an older file.

SolarDrew: I'm not sure this is true, because the unlink command is deliberately in the second except block so the file won't ever have been deleted if the first one catches.

dkist/net/attrs_values.py
```diff
@@ -98,42 +107,35 @@ def attempt_local_update(*, timeout: int = 1, user_file: Path = None, silence_er
...
-        if not silence_errors:
-            raise
-        success = False
 
-    return success
+        return True
+
+    except Exception as err:
+        log.error("Downloaded file could not be saved or is not valid JSON, removing cached file.")
+        user_file.unlink(missing_ok=True)
```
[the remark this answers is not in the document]
SolarDrew: And now having read the rest of the file I realise that is in fact what you're doing. :facepalm: As you were.

Cadair (approved):

PR #408, opened 2024-06-27
Plotting helpers for TiledDataset

### LINKED ISSUE #405: Add a `.plot` helper which produces the tiled plot in https://docs.dkist.nso.edu/projects/python-tools/en/stable/howto_guides/reproject_vbi_mosaic.html

SolarDrew:
Adds a `TiledDataset.plot()` method which creates a quick-look plot as per the example [here](https://docs.dkist.nso.edu/projects/python-tools/en/stable/howto_guides/reproject_vbi_mosaic.html).

At the moment only plots the first image in each tile. Unsure whether to try and construct an animation like the normal dataset plotting or to pass an index argument to the plot to determine which slice is shown.

Closes #405

### DIFF
--- a/dkist/dataset/tests/test_tiled_dataset.py
+++ b/dkist/dataset/tests/test_tiled_dataset.py
@@ -1,5 +1,6 @@
 import copy
 
+import matplotlib.pyplot as plt
 import numpy as np
 import pytest
 
@@ -59,3 +60,10 @@ def test_tiled_dataset_from_components(dataset):
         assert ds.files == fm
         assert ds.meta["inventory"] is inventory
         assert ds.meta["headers"] is headers
+
+
+@pytest.mark.mpl_image_compare
+@pytest.mark.parametrize("share_scale", [True, False])
+def test_plot(large_tiled_dataset, share_scale):
+    large_tiled_dataset.plot(0, share_scale=share_scale)
+    return plt.gcf()
--- a/dkist/dataset/tests/test_tiled_dataset.py
+++ b/dkist/dataset/tests/test_tiled_dataset.py
@@ -5,6 +5,7 @@ import numpy as np
 import pytest
 
 from dkist import Dataset, TiledDataset, load_dataset
+from dkist.tests.helpers import figure_test
 
 
 def test_tiled_dataset(simple_tiled_dataset, dataset):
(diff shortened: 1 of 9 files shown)


Cadair:
@SolarDrew can you rebase this horror show?

dkist/dataset/tests/test_tiled_dataset.py
```diff
@@ -59,3 +61,18 @@ def test_tiled_dataset_from_components(dataset):
         assert ds.files == fm
         assert ds.meta["inventory"] is inventory
         assert ds.meta["headers"] is headers
+
+
+# @pytest.mark.mpl_image_compare
```
Cadair, line 66:
```suggestion
```

dkist/dataset/tests/test_tiled_dataset.py
```diff
@@ -59,3 +61,18 @@ def test_tiled_dataset_from_components(dataset):
...
+@figure_test
+@pytest.mark.parametrize("share_scale", [True, False])
+def test_plot(share_scale):
```
Cadair, line 69:
```suggestion
@pytest.mark.parametrize("share_zscale", [True, False], ids=["share_zscale", "indpendent_zscale"])
def test_tileddataset_plot(share_zscale):
```

Cadair (commented):
Looks like codecov isn't uploading from the figure tests?

dkist/dataset/tests/test_tiled_dataset.py
SolarDrew: I thought this was redundant with `@figure_test`?

SolarDrew: Never mind I misread the suggestion. As you were.

SolarDrew:
I think this is good to go unless there are any more issues you can see with it.

dkist/dataset/tiled_dataset.py
```diff
@@ -124,5 +125,30 @@ def shape(self):
...
+            axmin, axmax = ax.get_images()[0].get_clim()
+            vmin = axmin if axmin < vmin else vmin
+            vmax = axmax if axmax > vmax else vmax
+            ax.set_ylabel(" ")
+            ax.set_xlabel(" ")
+        if share_zscale:
+            for ax in fig.get_axes():
+                ax.get_images()[0].set_clim(vmin, vmax)
+        timestamp = self[0, 0].axis_world_coords("time")[-1].iso[slice_index]
+        fig.suptitle(f"TiledDataset {self.inventory['datasetId']} at time {timestamp} (slice={slice_index})", y=0.95)
```
Cadair, line 151: I think it would be better if this were "VBI Dataset (ASKJD) at time ..." if you don't mind.

PR #437, opened 2024-09-13
Add FileManager to TiledDataset

### LINKED ISSUE #244: `TiledDataset` needs more helpers for working over all tiles
Maybe we need some kind of slicing helper and also the `TiledDataset` also probably needs it's own `FileManager`

SolarDrew:
Does what it says on the tin. Addresses but doesn't necessarily close #244


dkist/io/asdf/converters/tiled_dataset.py
```diff
@@ -11,7 +14,19 @@ class TiledDatasetConverter(Converter):
     def from_yaml_tree(cls, node, tag, ctx):
         from dkist.dataset.tiled_dataset import TiledDataset
 
-        return TiledDataset(node["datasets"], node["inventory"])
+        tiled_dataset = TiledDataset(node["datasets"], node["inventory"])
```
Cadair, line 17: Just a drive-by comment from my phone, but could we have this be built on demand in a property on TiledDataset rather than here?

With this implemented like this, does it reflect slicing of the sub-datasets correctly?

SolarDrew: I'm not sure if it does, that was the next thing I was going to test. But probably not. Presumably the aim of doing it as a property would be to have that automatically? I expect that would be easy to implement.

SolarDrew:
Well I'm slightly confused and concerned by the 10% speed improvement to `test_generate_celestial()` because I didn't think I'd changed anything that would affect it...

dkist/dataset/tiled_dataset.py
```diff
@@ -56,6 +59,8 @@ class TiledDataset(Collection):
 
     """
 
+    _file_manager = FileManagerDescriptor(default_type=FileManager)
```
Cadair, line 62: I don't think we need this if we are making `_file_manager` a property.

dkist/dataset/tiled_dataset.py
```diff
@@ -190,3 +193,24 @@ def __repr__(self):
...
+    def files(self):
+        """
+        A `~.FileManager` helper for interacting with the files backing the data in this ``Dataset``.
+        """
+        return self._file_manager
+
+    @property
+    def _file_manager(self):
+        return FileManager(
+            StripedExternalArray(
```
Cadair, line 207: I think here we need some more safeguards that all the tiles have the same settings in their respective file managers. target, dtype, shape, basepath, and I think even chunksize should all be pulled from the `FileManager` instance of the first tile and then verified that all the tiles are the same otherwise this should error.

Cadair (approved):

PR #439, opened 2024-09-23
Enable passing datasets to tests via command line

SolarDrew:
Adds a pytest cli option `--ds` which accepts a comma-separated list of strings that can be parsed by `load_dataset()`. If present, these are passed to any test with the `accept_cli_datasets` marker in place of the fixture dataset(s) actually specified for that test in the code.

The aim here is to be able to run arbitrary datasets through the test suite to more easily find issues with new data.


dkist/dataset/tests/test_dataset.py
```diff
@@ -43,20 +43,23 @@ def test_init_missing_meta_keys(identity_gwcs):
         Dataset(data, wcs=identity_gwcs, meta={"headers": {}})
 
 
+@pytest.mark.accept_cli_datasets
```
Cadair, line 46: I suggest we add a kwarg to this mark which tells it *which* argument to replace with the cli argument, so in this case it would be:

```suggestion
@pytest.mark.accept_cli_datasets(fixture="dataset")
```

we could make it default to "first argument" if we want, but this might be more explicit.

dkist/dataset/tests/test_tiled_dataset.py
```diff
@@ -28,6 +29,7 @@ def test_tiled_dataset_slice(simple_tiled_dataset, aslice):
     assert np.all(simple_tiled_dataset[aslice] == simple_tiled_dataset._data[aslice])
 
 
+@pytest.mark.accept_cli_datasets
 @pytest.mark.parametrize("aslice", [np.s_[0, :100, 100:200]])
 def test_tiled_dataset_slice_tiles(large_tiled_dataset, aslice):
```
Cadair, line 34: The issue here is that if you pass in a tiled dataset with the `--ds` arg then the tests expecting a dataset object will fail. We could add a `tiled=True` kwarg to `accept_cli_datasets` and a `--tiled-ds` cli arg to differentiate?

dkist/conftest.py
```diff
@@ -347,3 +347,14 @@ def visp_dataset_no_headers(tmp_path_factory):
...
+
+
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_call(item):
+    dslist = item.config.getoption("--ds")
+    dslist = dslist.split(",") if dslist else None
+    if dslist and item.get_closest_marker("accept_cli_datasets"):
+        dsargs = [load_dataset(ds) for ds in dslist]
+        for ds, arg in zip(dsargs, item.fixturenames):
+            item.funcargs[arg] = ds
```
Cadair, line 359: This is letting you pass in multiple datasets which get passed as multiple *different* fixtures to the tests, or passed in as multiple parametrized inputs to a single fixture? I feel like the second option is the most useful?

Cadair (commented):
This is neat, and will be useful. Needs documenting somewhere, though god knows where.

dkist/dataset/tests/test_dataset.py
SolarDrew: Good call, I was wondering how to make it more robust in cases where there are multiple fixtures.

dkist/conftest.py
SolarDrew: It's passing them in as different fixtures, although I'm not sure that as written it will actually work for eg `some_test(fixture_a, fixture_b`) if `fixture_a` depends on other fixtures, because of how pytest orders them in the list of arguments. Your suggestion of modifying the marker should fix that though.

I think passing in parametrized inputs to a single fixture could be useful but I think that's a different task, unless I've misunderstood you.

Cadair: Do we need to support more than one fixture per test? surely almost all of them only use one?

SolarDrew: Almost all do, yes, but a few use two, for example to compare more complex datasets to the base one. Those ones probably do want to be reworked a bit with this anyway so it may be that we can refactor out needing more than one fixture input.

Cadair: Do we need to support that use case here? feels too niche?

SolarDrew: Probably not, I'm happy to just leave it

dkist/dataset/tests/test_tiled_dataset.py
Cadair: An alternate thought, instead of `tiled=true` and `fixture=` we could have a `dataset="fixture_name"` and `tiled_dataset="fixture_name"` pair of kwargs?

dkist/dataset/tests/test_dataset.py
```diff
@@ -51,13 +52,15 @@ def test_repr(dataset, dataset_3d):
 
 
 def test_wcs_roundtrip(dataset):
+    # TOO generalise this so mark.accept_cli_dataset will work, if possible
     p = (10*u.pixel, 10*u.pixel)
```
Cadair, line 56: This should do the trick:
```suggestion
    p = [1*u.pixel] * dataset.wcs.pixel_n_dim
```

dkist/dataset/tests/test_dataset.py
```diff
@@ -51,13 +52,15 @@ def test_repr(dataset, dataset_3d):
...
     w = dataset.wcs.pixel_to_world(*p)
     p2 = dataset.wcs.world_to_pixel(w)
     assert_quantity_allclose(p, p2 * u.pix)
 
 
 def test_wcs_roundtrip_3d(dataset_3d):
+    # TOO generalise this so mark.accept_cli_dataset will work, if possible
```
Cadair, line 63: I don't know what this test is doing, but if we generalise the one above this one can stay here and be weird by itself.

```suggestion
```

docs/developer.rst
```diff
@@ -92,3 +92,25 @@ Sphinx References
...
+-------
+
+Testing with arbitrary datasets
+*******************************
+
+The DKIST Python tools allow the test suite to run on user-provided datasets.
+To enable this there are two pytest command line optionsa available, `--ds` and `--tiled-ds`.
+Each takes a string parseable by `load_dataset()`.
+Tests with the `accept_cli_dataset` or `accept_tiled_dataset` markers will take the value given for `--ds` or `--tiled-ds` respectively, load that dataset and pass it into the test in place of a fixture.
+By default the first fixture specified in the test definition is the one replaced, but if a fixture name is given to the marker that fixture will be replaced instead.
```
Cadair, line 106:
```suggestion
To only run tests using the supplied datasets you can run pytest with `-m accept_cli_dataset` and/or `-m accept_cli_tiled_dataset`.
By default the first fixture specified in the test definition is the one replaced, but if a fixture name is given to the marker that fixture will be replaced instead.
```

Cadair (approved):

dkist/dataset/tests/test_dataset.py
SolarDrew: That makes sense but then do we still need to have a 2d and 3d version of this test? Or shall we just lump them together with this change?

SolarDrew: Oh you've already commented on that. Is the other one not doing the same as this but in 3d?

Cadair: I don't know why the other test splits the pixel asserts, we can try it and see though.

SolarDrew: Ah I assume it's because the third dimension has different dimensions, but we can probably account for that as well.

SolarDrew: > This should do the trick:

Turns out it does not because the order `pixel_to_world` gives coords in is not necessarily the order `world_to_pixel` needs to take them in. This probably can be generalised and it may not even be too hard but I think it's not worth it blocking this any further right now.

Cadair: errr, wat? I think that is a bug

SolarDrew: For the record, per conversation offline yes this is a bug but it's a bug in the gwcs and not in the tests, so the test can stay as it is, since catching these bugs is exactly the point.

Cadair:
Thanks Drew

PR #440, opened 2024-09-26
Correct bug in #439

SolarDrew:
Small error which slipped past me.


Cadair:
Should we have a tox test env where we pass in the sample data via the cli to see what happens?

Cadair (approved):

SolarDrew:
Yeah I think that's a good idea actually. Wonder if that could be hooked into something to automate a list of known failures :thinking:

PR #441, opened 2024-09-30
Update `TiledDataset.plot()`

SolarDrew:
Some tweaks to TiledDataset plotting to work for DL-NIRSP data, or hopefully arbitrary arrangements of tiled data.

### DIFF
--- a/dkist/dataset/tiled_dataset.py
+++ b/dkist/dataset/tiled_dataset.py
@@ -152,7 +152,19 @@ class TiledDataset(Collection):
         """
         return [[tile.data.shape for tile in row] for row in self]
 
-    def plot(self, slice_index: int, share_zscale=False, **kwargs):
+    def plot(self, slice_index, share_zscale=False, **kwargs):
+        """
+        Plot a slice of each tile in the TiledDataset
+
+        Parameters
+        ----------
+        slice_index : `int`, sequence of `int`s or `numpy.s_`
+            Object representing a slice which will reduce each component dataset of the TiledDataset to a 2D image.
+        share_zscale : `bool`
+            Determines whether the color scale of the plots should be calculated independently (``False``) or shared across all plots (``True``).
+            Defaults to False
+
+        """
         if isinstance(slice_index, int):
             slice_index = (slice_index,)
         vmin, vmax = np.inf, 0
@@ -181,7 +193,7 @@ class TiledDataset(Collection):
             if coord == "time":
                 val = val.iso
             title += f"{coord} {val}" + (", " if i != len(slice_index)-1 else " ")
-        title += f"(slice={(slice_index if len(slice_index) > 1 else slice_index[0])})"
+        title += f"(slice={(slice_index if len(slice_index) > 1 else slice_index[0])})".replace("slice(None, None, None)", ":")
         fig.suptitle(title, y=0.95)
         return fig
 
--- a/dkist/dataset/tiled_dataset.py
+++ b/dkist/dataset/tiled_dataset.py
@@ -214,7 +214,7 @@ class TiledDataset(Collection):
         <dkist.dataset.tiled_dataset.TiledDataset object at 0x7024f7382cd0>
         This VBI Dataset AJQWW is an array of (3, 3) Dataset objects and
         consists of 9 frames.
-        Files are stored in /...
+        Files are stored in ...
 
         Each Dataset has 2 pixel and 2 world dimensions.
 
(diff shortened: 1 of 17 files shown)


dkist/dataset/tiled_dataset.py
```diff
@@ -153,10 +153,12 @@ def tiles_shape(self):
         return [[tile.data.shape for tile in row] for row in self]
 
     def plot(self, slice_index: int, share_zscale=False, **kwargs):
```
Cadair, line 155: What is `slice_index` now? is it actually a generic slice which should lead to a spatial-spatial plot?

Also this method needs a docstring.

SolarDrew: It's an int or tuple that specifies the slice of each tile to plot, ie it's equivalent to the input of `slice_tiles`. That will give a spatial-spatial plot if those are the last two pixel dimensions, but obviously that's not guaranteed. I guess in principle if I generalise it to take an actual slice object as well then it will be able to plot whatever arrangement of coords you want, but I'm not sure how much physical sense that makes for multi-dimensional tiles.

Cadair: I think this would work if you did give it a full output of `np.s_`? I think this probably needs documenting well and then testing a bunch so we understand any quirks here.

dkist/dataset/tiled_dataset.py
```diff
@@ -152,11 +152,25 @@ def tiles_shape(self):
...
         return [[tile.data.shape for tile in row] for row in self]
 
-    def plot(self, slice_index: int, share_zscale=False, **kwargs):
+    def plot(self, slice_index, share_zscale=False, **kwargs):
+        """
+        Plot a slice of each tile in the TiledDataset
+
+        Parameters
+        ----------
+        slice_index : `int`, sequence of `int`s or `numpy.s_`
```
Cadair, line 161: Is this the same as the input to `slice_tiles`? if yes it should be referenced from here. Also that needs a docstring now I look at it.

dkist/dataset/tiled_dataset.py
```diff
@@ -174,12 +188,20 @@ def plot(self, slice_index: int, share_zscale=False, **kwargs):
...
+                val = val.iso
+            title += f"{coord} {val}" + (", " if i != len(slice_index)-1 else " ")
+        title += f"(slice={(slice_index if len(slice_index) > 1 else slice_index[0])})".replace("slice(None, None, None)", ":")
+        fig.suptitle(title, y=0.95)
         return fig
 
     @property
     def slice_tiles(self):
+        """
+        Returns a new TiledDataset with the given slice applied to each of the tiles.
```
Cadair, line 203: I think this needs an example and some more info. It's not obvious from this what the syntax is.

dkist/dataset/tiled_dataset.py
```diff
@@ -174,12 +189,56 @@ def plot(self, slice_index: int, share_zscale=False, **kwargs):
...
+
+        Examples
+        --------
+        >>> from dkist import load_dataset
+        >>> from dkist.data.sample import VBI_AJQWW
+        >>> ds = load_dataset(VBI_AJQWW)
+        >>> ds.slice_tiles[0, 10:-10]
```
Cadair, line 213:
```suggestion
        >>> from dkist.data.sample import VBI_AJQWW  # doctest: +REMOTE_DATA
        >>> ds = load_dataset(VBI_AJQWW)  # doctest: +REMOTE_DATA
        >>> ds.slice_tiles[0, 10:-10]  # doctest: +REMOTE_DATA
```

dkist/dataset/tiled_dataset.py
```diff
@@ -174,12 +189,56 @@ def plot(self, slice_index: int, share_zscale=False, **kwargs):
...
+        """
+        Returns a new TiledDataset with the given slice applied to each of the tiles.
+
+        Examples
+        --------
+        >>> from dkist import load_dataset
+        >>> from dkist.data.sample import VBI_AJQWW  # doctest: +REMOTE_DATA
+        >>> ds = load_dataset(VBI_AJQWW)  # doctest: +REMOTE_DATA
+        >>> ds.slice_tiles[0, 10:-10]  # doctest: +REMOTE_DATA
+        <dkist.dataset.tiled_dataset.TiledDataset object at 0x7024f7382cd0>
```
Cadair, line 214:
```suggestion
        <dkist.dataset.tiled_dataset.TiledDataset object at ...>
```

PR #447, opened 2024-10-14
Try to fix failing build

SolarDrew:


SolarDrew:
This looks like it's working now and seems to be just dependent on the fix for the figure-devdeps run.

Cadair (approved):

Cadair:
The figure is a known issue, so merged. Thanks @SolarDrew

PR #448, opened 2024-10-17
Add a devdeps figure build

Cadair:
We should test our figures against at least astropy main

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -46,12 +46,13 @@ deps =
     devdeps: git+https://github.com/spacetelescope/gwcs
     devdeps: git+https://github.com/asdf-format/asdf
     devdeps: git+https://github.com/astropy/asdf-astropy
+    devdeps: git+https://github.com/sunpy/mpl-animators
     # Autogenerate oldest dependencies from info in setup.cfg
     oldestdeps: minimum_dependencies
-    figure: matplotlib==3.9.1
-    figure: mpl_animators==1.1.1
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -46,12 +46,13 @@ deps =
...
     # Autogenerate oldest dependencies from info in setup.cfg
     oldestdeps: minimum_dependencies
-    figure: matplotlib==3.9.1
-    figure: mpl_animators==1.1.1
-    figure: astropy==6.1.1
-    figure: ndcube==2.2.2
+    figure-!devdeps: matplotlib==3.9.1
+    figure-!devdeps: mpl-animators==1.1.1
+    figure-!devdeps: astropy==6.1.1
+    figure-!devdeps: ndcube==2.2.2
```
SolarDrew, line 55: Should these packages not be pinned for `figure-devdeps` as well?

Cadair: We install all the git versions of them instead, so we can't have both.

SolarDrew: Oh yes, I see. Well then won't that (potentially) lead to figure differences as the packages change on git?

Cadair: yeah some of the differences could be something irrelevant has changed upstream, or it could be a legitimate API change. Unfortunately there's no way to test the latter without running the risk of the former. fwiw, random hash breakages don't seem to happen that much on astropy/sunpy.

SolarDrew (approved):

PR #454, opened 2024-10-24
Simplify axis label handling for TiledDataset.plot

Cadair:
This also means that we have compatiblity with https://github.com/astropy/astropy/pull/17243


Cadair:
This requires https://github.com/astropy/astropy/pull/17202 to work, so we would need to have a min astropy requirement of 6.1.5 before we can apply this patch.

PR #455, opened 2024-10-26
Speed up distributed Dask by removing reference to array_container in FITSLoader

svank:
I've been finding that the Dask arrays generated by this package don't give me the speed I would hope for, and this PR helps address that. (I'm working on another, too.)

My use case right now (my first with DKIST) is taking a ViSP data set and summing over wavelength to get a quasi-broadband view of what ViSP was seeing.

Doing this serially takes 1 m 53 s
```python
ds = dkist.load_dataset("/home/svankooten/BXXDZ/")

def comp_slice(file):
    data = fits.getdata("/home/svankooten/BXXDZ/" + file)
    data = data[0]
    return data.sum(axis=0)

image = np.stack(list(map(comp_slice, ds.files.filenames)))
```

Doing this with multiprocessing takes 16 s with 8 cores
```python
from multiprocessing import Pool

with Pool(8) as p:
    image = np.stack(p.map(comp_slice, ds.files.filenames))
```

The default dask scheduler uses threads, and with the GIL I get 250-300% CPU usage and it takes 1m 5s
```python
ds.data.sum(axis=-2).compute();
```

Switching to Dask's distributed scheduler evades the GIL and ideally will look like the multiprocessing case, but it actually takes longer than the serial case, at 2m 33s
```python
from dask.distributed import Client
client = Client(threads_per_worker=1, n_workers=8)
ds.data.sum(axis=-2).compute()
```

The task stream in the scheduler shows that there are lots of gaps between tasks
image

What I finally found as the cause is that each `AstropyFITSLoader` keeps a reference to its parent `StripedExternalArray`. As I understand it, the loaders get pickled and sent to the worker processes before they're used, and this reference causes these pickle payloads to be several megabytes each. This PR avoids pickling the `StripedExternalArray`. This makes the task stream much denser, and lets the computation complete in 47 s
image
(The x scale is different in this screenshot)

This beats the serial case (!) and the default threaded scheduler, but doesn't match multiprocessing. (Not sure how close this can get, I assume Dask has more overhead.) This PR doesn't change the run time for the default scheduler, which I suppose doesn't need to do any pickling.

This PR has the `Loaders` just record the `basepath` when they're created. If `Filemanager.basepath` is updated later on, the setter in `StripedExternalArray` now iterates through `loader_array` and updates each one. This has the downside of no longer having a single source of truth for `basepath`, but I think it's worth the speedup. (And it seems like a user would have to dig deep in internal API to change one but not the others.)

An alternate design I tried was to have have `BaseFITSLoader` keep its `array_container` reference, but I overrode `BaseFITSLoader`'s pickling behavior so that only `array_container.basepath` is sent, not `array_container` itself. Then the unpickled loaders in the worker processes set a `_basepath` value (a snapshot of the "live" value at the time of pickling, which is when the dask calculation is triggered) but don't set `array_container`, and the `basepath` property returns whichever is set. This lets `basepath` be stored in only one location in the main process, but I think it has the side-effect that if a user loads a dataset, pickles it for some reason, unpickles it, updates the basepath and then triggers the Dask calculation, the pickling will have caused the loaders to drop their references and they won't get the updated basepath.

Hopefully this PR doesn't conflict with any future plans for `FITSLoader`!

### DIFF
--- a/dkist/io/file_manager.py
+++ b/dkist/io/file_manager.py
@@ -97,14 +97,13 @@ class StripedExternalArray(BaseStripedExternalArray):
         self.dtype = dtype
         self.target = target
         self._loader = loader
-        self._basepath = None
-        self.basepath = basepath  # Use the setter to convert to a Path
+        self._basepath = self._sanitize_basepath(basepath)
         self.chunksize = chunksize
         self._fileuri_array = np.atleast_1d(np.array(fileuris))
 
         loader_array = np.empty_like(self._fileuri_array, dtype=object)
         for i, fileuri in enumerate(self._fileuri_array.flat):
-            loader_array.flat[i] = loader(fileuri, shape, dtype, target, self)
+            loader_array.flat[i] = loader(fileuri, shape, dtype, target, self.basepath)
 
         self._loader_array = loader_array
 
@@ -125,13 +124,15 @@ class StripedExternalArray(BaseStripedExternalArray):
         """
         return self._basepath
 
+    @staticmethod
+    def _sanitize_basepath(value):
+        return Path(value).expanduser() if value is not None else None
+
     @basepath.setter
     def basepath(self, value: os.PathLike | str | None):
-        self._basepath = Path(value).expanduser() if value is not None else None
-        if hasattr(self, '_loader_array'):
-            # We get called during __init__ before _loader_array is created
-            for loader in self._loader_array.flat:
-                loader.basepath = self._basepath
+        self._basepath = self._sanitize_basepath(value)
+        for loader in self._loader_array.flat:
+            loader.basepath = self._basepath
 
     @property
     def fileuri_array(self) -> NDArray[str]:
--- a/dkist/io/file_manager.py
+++ b/dkist/io/file_manager.py
@@ -128,6 +128,10 @@ class StripedExternalArray(BaseStripedExternalArray):
     @basepath.setter
     def basepath(self, value: os.PathLike | str | None):
         self._basepath = Path(value).expanduser() if value is not None else None
+        if hasattr(self, '_loader_array'):
+            # We get called during __init__ before _loader_array is created
+            for loader in self._loader_array.flat:
+                loader.basepath = self._basepath
 
     @property
     def fileuri_array(self) -> NDArray[str]:
--- a/dkist/io/loaders.py
+++ b/dkist/io/loaders.py
@@ -50,7 +50,7 @@ class BaseFITSLoader(metaclass=abc.ABCMeta):
         self.shape = shape
         self.dtype = dtype
         self.target = target
-        self.array_container = array_container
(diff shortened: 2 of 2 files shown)


Cadair:
Thanks a lot for your PR @svank :rocket: 

This is a great change, we haven't really had the time to do any significant optimisation on the loading code, so anything you can find would be very welcome.

I was curious to see the comparison between using distributed and using the [local processes](https://docs.dask.org/en/stable/scheduling.html#local-processes) scheduler, so I checked out your branch and loaded your data.

<details>
<summary>Setup</summary>

```python
import dask
from multiprocessing import Pool
import dkist
from astropy.io import fits
import numpy as np

ds = dkist.load_dataset("~/dkist_data/pid_2_86/BXXDZ")

ds_rebin = ds.rebin((1, 1, 936, 1))

def comp_slice(file):
    data = fits.getdata(ds.files.basepath / file)
    data = data[0]
    return data.sum(axis=0)
```
</details>

```python
# baseline serial, numpy
%timeit -n 1 -r 1 np.stack(list(map(comp_slice, ds.files.filenames)))
# 1min 57s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
```

```python
# simple multi-processing
p = Pool(16)
%timeit -n 1 -r 1 np.stack(p.map(comp_slice, ds.files.filenames))
# 12.3 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
p.close()
```
```python
# Processes

dask.config.set(scheduler='processes')

%timeit -n 1 -r 1 ds.data.sum(axis=-2).compute();
# 18.1 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)

%timeit -n 1 -r 1 ds_rebin.data.compute()
# 18.3 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
```

it's quite close, certainly good enough imo.

Without this PR, the last example took: `4min 34s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)` so this looks to be a massive improvement to me.

svank:
Interesting to see! With this PR I also have the `processes` scheduler coming in a bit slower than DIY parallelism, but twice as fast as the distributed scheduler. I wonder why---distributed's task stream has gaps but it's still mostly full. Maybe it has more overhead around starting and stopping each task?

Cadair:
I suspect there's more overhead, but I don't know the details. 2x as fast without distributed is certainly eyebrow raising.

Cadair:
@astrofrog suggested to me that it might be something to do with differences in how local process and distributed do or do not use threads on each worker.

astrofrog:
As suggested to @Cadair right now I think the issue is that the chunks are very small and take <100ms to run, so distributed is going to be inefficient. I think it would be worth trying calling ``.rechunk`` to see if this helps.

PR #459, opened 2024-10-29
Remove pin on pytest-benchmark

Cadair:


Cadair:
It seems that the issue has been fixed.

PR #462, opened 2024-11-05
Making the array

svank:
I'm unsure if this PR is worth it, but it offers some extra analysis on the optimization/chunk size front.

This PR speeds up computation with the distributed scheduler a bit by changing the code for building the Dask array to simplify the task graph (or so I believe). Instead of several iterations of `da.stack` calls, it directly generates Dask tasks for each chunk of the final array, following the [example here](https://docs.dask.org/en/latest/array-design.html). This shaves a bit of time off when using the distributed scheduler (36 to 32 seconds), but the "processes" scheduler is still faster and is almost unaffected by this change (from a consistent 20.5 to 19.7 seconds). I'm using the same test case as in #455 .

This PR currently needs more thought on handling non-default chunksizes (the argument to `stack_loader_array`, which it looks like can be set in the ASDF file?). I guess I'm not sure what the intended use case is for that. In `main` it looks like this option can only sub-divide files into smaller chunks. Are there DKIST files out there that are too big to be a single chunk? Or is the idea to subdivide files when only a portion of each file has to be loaded? As this PR is now, each file gets fully loaded, and then split up, which might be less efficient. But before this PR, I'm not sure if that was also the case, or if `hdu.data[slice]` loads just the sliced portion of the file. If the latter, changing this PR to make use of that would, I think, require manually creating tasks for each chunk of each file, which seems complicated.

In #455, there was discussion of rechunking to larger chunk sizes. At least for that test case, using `.rechunk((1, 100, -1, -1))` actually slows down the total computation (I'm seeing it go from 36 to 45 seconds), I assume from the overhead of gathering and combining arrays. This PR's design makes it easy to try larger chunk sizes more efficiently, as one "loading" task can load multiple files and concatenate them immediately, rather than adding a separate layer to the task graph. This is implemented in this PR, in the `Docstring` commit and before. A "batch size" can be set, and each Dask task loads that many files and concatenates them. (The batch size has to be adjusted in the code---I didn't know how to expose it as an option for `load_dataset`.) I found that with a batch size of 100 (producing chunks of a few hundred MB, Dask's recommended size), I was actually getting a very slightly longer runtime (going from 24 to 27 seconds), but it's less extreme than using `.rechunk()`. My guess is that the `np.concatenate` call copying all the data arrays offsets any efficiency gains from the larger chunks. (Maybe if I were doing a lot more with the data it would be worth it.) I didn't see a way to have Astropy FITS load a file directly into a view of a larger output array, unfortunately.

Here's the task stream (top is before, bottom is this PR):
image

And here's a portion of the task graph for each (left is before, right is this PR). This is from `dataset.data.visualize()`, and I made an asdf file with only a handful of FITS files so the images wouldn't be unrenderably large.
image

Right now my thinking is "just use the `processes` scheduler", where the gains from this PR are small enough that I'm not sure it's worth more work to handle the `chunksize` argument better.

### DIFF
--- a/dkist/io/dask_utils.py
+++ b/dkist/io/dask_utils.py
@@ -6,7 +6,7 @@ import numpy as np
 __all__ = ["stack_loader_array"]
 
 
-def stack_loader_array(loader_array, chunksize=None, batch_size=1):
+def stack_loader_array(loader_array, output_shape, chunksize=None, batch_size=1):
     """
     Stack a loader array along each of its dimensions.
 
@@ -35,11 +35,11 @@ def stack_loader_array(loader_array, chunksize=None, batch_size=1):
                              name="load_files",
                              chunks=chunks,
                              dtype=loader_array.flat[0].dtype)
+    array = array.reshape(output_shape)
     if chunksize is not None:
         # If requested, re-chunk the array. Not sure this is optimal
         new_chunks = (1,) * (array.ndim - len(chunksize)) + chunksize
         array = array.rechunk(new_chunks)
-    # The calling function reshapes the array to the final, correct shape
     return array
 
 
--- a/dkist/io/dask_utils.py
+++ b/dkist/io/dask_utils.py
@@ -1,12 +1,9 @@
-from itertools import batched
-
 import dask
-import numpy as np
 
 __all__ = ["stack_loader_array"]
 
 
-def stack_loader_array(loader_array, output_shape, chunksize=None, batch_size=1):
+def stack_loader_array(loader_array, output_shape, chunksize=None):
     """
     Converts an array of loaders to a dask array that loads a chunk from each loader
 
@@ -20,8 +17,6 @@ def stack_loader_array(loader_array, output_shape, chunksize=None, batch_size=1)
         The intended shape of the final array
     chunksize : tuple[int]
         Can be used to set a chunk size. If not provided, each batch is one chunk
-    batch_size : int
-        The number of files to load in each dask task
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


Cadair:
> I guess I'm not sure what the intended use case is for that. In main it looks like this option can only sub-divide files into smaller chunks. Are there DKIST files out there that are too big to be a single chunk? Or is the idea to subdivide files when only a portion of each file has to be loaded?

This was the idea yes, although as time has progressed I don't think it's a good idea. In theory we could chunk down to the individual compressed tiled in the FITS file, but that would make the dask graph insanely large with insanely small chunk sizes so I don't think we are likely to do it.

dkist/io/dask_utils.py
```diff
@@ -1,60 +1,57 @@
...
+    for i, loader in enumerate(loader_array.flat):
+        # The key identifies this chunk's position in the (partially-flattened) final data cube
+        key = ("load_files", i)
+        key += (0,) * len(file_shape)
+        # Each task will be to call _call_loader, with the loader as an argument
+        tasks[key] = (_call_loader, loader)
+
+    dsk = dask.highlevelgraph.HighLevelGraph.from_collections("load_files", tasks, dependencies=())
+    # Specifies that each chunk occupies a space of 1 pixel in the first dimension, and all the pixels in the others
+    chunks = ((1,) * loader_array.size,) + tuple((s,) for s in file_shape)
```
Cadair, line 38: I think this is only true for some arrays? i.e. VISP and not VBI?

Cadair:
@svank If we agreed to get rid of the sub-file loading chunksize (I think this would be dropping doing anything with chunksize completely) would you be willing to get this PR mergeable?

I like the idea of being able to batch chunks, did you try out that functionality with the distributed scheduler? I wonder if it would make more difference there?

dkist/io/dask_utils.py
svank: I think at this stage, the array is just `(n_chunks, *chunk_size)`, and a few lines down the actual data cube shape is imposed. I think that approach seemed easier that figuring out how to assemble the loaders into the actual data cube shape from the beginning.

Cadair:
I've kept chunksize around but made it throw a deprecation warning (which I don't expect people to ever see).

pytest.ini
```diff
@@ -74,3 +74,4 @@ filterwarnings =
     ignore:pkg_resources is deprecated as an API.*:DeprecationWarning
     ignore:Deprecated call to .*pkg_resources\.declare_namespace.*mpl_toolkits.*:DeprecationWarning
     ignore:Signature b:UserWarning
+	ignore:Reshaping is producing a large chunk:dask.array.core.PerformanceWarning
```
Cadair, line 77:
```suggestion
    ignore:Reshaping is producing a large chunk:dask.array.core.PerformanceWarning
```

SolarDrew (approved):

PR #476, opened 2024-12-09
Pin dask to <2024.12.0 to avoid performance regression

SolarDrew:


SolarDrew:
Yeah that's fair. I'm not sure we want to take the performance hit either though, it's not small.

Cadair:
Yeah, we should debug it further and figure it out. This seem to have not fixed it anyway?

SolarDrew:
Presumably that's because it's running the unpinned version as it exists on main?

SolarDrew:
Closing until we've figured out what is actually happening.

PR #487, opened 2025-01-14
Add support to TiledDataset for missing, irregular or overlapping tiles

SolarDrew:
Changes the array of datasets underlying TiledDataset into a masked array, defaulting to False everywhere (all tiles are valid).


dkist/dataset/tiled_dataset.py
```diff
@@ -82,8 +82,8 @@ def _from_components(cls, shape, file_managers, wcses, header_tables, inventory)
 
         return cls(datasets, inventory)
 
-    def __init__(self, dataset_array, inventory=None):
-        self._data = np.array(dataset_array, dtype=object)
+    def __init__(self, dataset_array, inventory=None, mask=False):
+        self._data = np.ma.masked_array(dataset_array, dtype=object, mask=mask)
```
Cadair, line 86: Can we auto-generate the mask based on the elements of dataset_array?

SolarDrew: I expect so. We can readily assume anything that isn't a Dataset should be masked out, but we should probably also have some way of flagging invalid Datasets for masking as well.

dkist/io/asdf/converters/tiled_dataset.py
```diff
@@ -5,16 +5,23 @@ class TiledDatasetConverter(Converter):
...
     types = ["dkist.dataset.tiled_dataset.TiledDataset"]
 
     def from_yaml_tree(cls, node, tag, ctx):
         from dkist.dataset.tiled_dataset import TiledDataset
 
-        return TiledDataset(node["datasets"], node["inventory"])
+        try:
+            mask = node["mask"]
+        except KeyError:
+            mask = None
```
Cadair, line 18:
```suggestion
        mask = node.get("mask", None)
```

dkist/io/asdf/converters/tiled_dataset.py
```diff
@@ -5,16 +5,23 @@ class TiledDatasetConverter(Converter):
...
+
+        return TiledDataset(node["datasets"], node["inventory"], mask)
 
     def to_yaml_tree(cls, tiled_dataset, tag, ctx):
         tree = {}
         tree["inventory"] = tiled_dataset._inventory
         tree["datasets"] = tiled_dataset._data.tolist()
+        tree["mask"] = tiled_dataset._data.mask.tolist()
```
Cadair, line 26: I worry that doing this will add more yaml to the file, when most of the time we probably don't need it, i.e. it's the same as "where datasets is not None". I am not sure if it's worth even trying to investigate this or not.

dkist/io/asdf/resources/schemas/tiled_dataset-1.1.0.yaml
Cadair, line 1: Because you added a new schema (which is correct), I think there might be more places you need to add the id of this schema, like a resources file? You might need a whole new manifest as well.

SolarDrew:
There are a few places where this can still be improved - if we're happy to make issues for those and punt them then this is good to go pending review. Otherwise I'll come back to it on Monday.

dkist/dataset/tiled_dataset.py
Cadair: Shouldn't the default value of mask be `None`?

dkist/io/asdf/converters/tiled_dataset.py
Cadair: Following more discussion in #500 I think we should just make this a numpy array and not a list.

Cadair (commented):
:+1: This looks good.

I think we should postpone merging this until we have made the corresponding changes to dkist-inventory just to make sure we haven't missed anything.

dkist/dataset/tiled_dataset.py
SolarDrew: Either works fine. `False` is specifying  to set the mask to be `False` (ie: not masked) everywhere. Which is the default if you pass `None` into `masked_array` anyway, but explicit is better than implicit.

Cadair: If we set it to `None` that could be "auto" where as `False` would be no-mask.

dkist/io/asdf/resources/schemas/tiled_dataset-1.2.0.yaml
```diff
@@ -0,0 +1,45 @@
...
+      inventory:
+        description: A copy of the inventory record for this dataset.
+        type: object
+
+    required: [inventory]
+    additionalProperties: true
+
+  mask:
+    description: A mask to indicate if invalid or missing Datasets should be ignored.
+    datatype: bool8
```
Cadair, line 41: I think this needs to be a fuller ndarray type?
```suggestion
    tag: tag:stsci.edu:asdf/core/ndarray-1.*
    datatype: bool8
    exact_datatype: true
    ndim: 2
```

Cadair (changes requested):
Needs a figure test for masked (and actually missing) tiles

dkist/dataset/tiled_dataset.py
```diff
@@ -37,10 +37,13 @@ def __init__(self, data, meta):
         self.meta = meta
 
     def __getitem__(self, slice_):
-        new_data = []
-        for tile in self.data.flat:
-            new_data.append(tile[slice_])
-        return TiledDataset(np.array(new_data).reshape(self.data.shape), meta=self.meta)
+        new_data = np.ma.zeros_like(data)
+
+        for i, ds in enumerate(data.flat):
```
Cadair, line 42:
```suggestion
        new_data = np.ma.zeros_like(self._data)

        for i, ds in enumerate(self._data.flat):
```

dkist/dataset/tiled_dataset.py
```diff
@@ -337,29 +371,5 @@ def files(self):
...
-                fileuris=fileuris,
-                target=1,
-                dtype=dtype,
-                shape=shape,
-                loader=AstropyFITSLoader,
-                basepath=basepath,
-                chunksize=chunksize
-            )
-        )
+
```
Cadair, line 374:
```suggestion
```

Cadair:
This is close. It needs:

* doc build fixing
* Well written changelogs detailing all the changes, the breakage isn't likely to be common API but it's real.
* Tests for `Tileddataset.files.[download | basepath.setter]` etc.
* Needs tests, including figure tests, for masked-but-present and masked-and-missing datasets.

docs/reference.rst
```diff
@@ -22,7 +22,7 @@ Reference / API
    :headings: ^#
    :no-inheritance-diagram:
 
-.. automodapi:: dkist.io.loaders
+.. automodapi:: dkist.io.dask.loaders
```
Cadair, line 25: This should probably be all of the new `.dask` submodule. (Either all the sub-submodules or we need to export stuff up into the `io.dask` namespace).

SolarDrew:
>     * Tests for `Tileddataset.files.[download | basepath.setter]` etc.

Does [this](https://github.com/DKISTDC/dkist/pull/487/files#diff-4c0bc6fe9f43636b20a363dcc20a142ccfb22a1e9d19738821f3a82b1e1bdf89R197) not test the latter?

SolarDrew:
> > ```
> > * Tests for `Tileddataset.files.[download | basepath.setter]` etc.
> > ```
> 
> Does [this](https://github.com/DKISTDC/dkist/pull/487/files#diff-4c0bc6fe9f43636b20a363dcc20a142ccfb22a1e9d19738821f3a82b1e1bdf89R197) not test the latter?

Apparently not because those lines aren't getting hit.

dkist/conftest.py
```diff
@@ -299,23 +299,32 @@ def eit_dataset():
...
+                         [False, False, False]]],
+                ids=["large-nomask", "large-masked"])
+def large_tiled_dataset(tmp_path_factory, request):
     vbidir = tmp_path_factory.mktemp("data")
     with gzip.open(Path(rootdir) / "large_vbi.asdf.gz", mode="rb") as gfo:
         with open(vbidir / "test_vbi.asdf", mode="wb") as afo:
             afo.write(gfo.read())
-    return load_dataset(vbidir / "test_vbi.asdf")
+    ds = load_dataset(vbidir / "test_vbi.asdf")
+    ds._data.mask = request.param
```
Cadair, line 326:
```suggestion
    ds.mask = request.param
```

dkist/io/asdf/converters/tiled_dataset.py
```diff
@@ -29,4 +31,5 @@ def to_yaml_tree(cls, tiled_dataset, tag, ctx):
         meta.pop("history", None)
         tree["meta"] = meta
         tree["datasets"] = tiled_dataset._data.tolist()
+        tree["mask"] = tiled_dataset._data.mask
```
Cadair, line 34:
```suggestion
        tree["mask"] = tiled_dataset.mask
```

Cadair (approved):

PR #489, opened 2025-01-14
Update transfer.py with as_posix paths

krislarsnso:
Works on my Windows machine.  I have not tested it on my Ubuntu emulator, but it should work.  Thanks to @Cadair for help with this.

### DIFF
--- a/changelog/489.bugfix.rst
+++ b/changelog/489.bugfix.rst
@@ -1,0 +1,1 @@
+Fix path errors with Globus transfers on windows.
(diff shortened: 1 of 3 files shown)


Cadair:
Thanks @krislarsnso I'd like to figure out how to unit test this, but I am happy to look at that if you have other things to worry about 🤣

Cadair:
Turns out we did have unit tests and they were just incorrect. The absolute best kind of tests where they have the same bug as the code.

PR #503, opened 2025-01-22
Fixing loading ASDF files from a directory

### LINKED ISSUE #496: Update dataset loader to handle the potential of many asdf files for a single dataset due to naming suffix changes
Over the last while the data centre has made ASDF files with two different naming conventions:

- `f"{instrument}_L1_{start_time:%Y%m%dT%H%M%S}_{dataset_id}..asdf"`
- `f"{instrument}_L1_{start_time:%Y%m%dT%H%M%S}_{dataset_id}_user_tools..asdf"`

In the future this suffix will be changed to:

- `f"{instrument}_L1_{start_time:%Y%m%dT%H%M%S}_{dataset_id}_metadata.asdf"`

Therefore we need to adapt the directory loader to handle picking the most applicable of these patterns if more than one is detected in the folder. I think the algorithm should be as follows:

- Glob the directory for all ASDF files
- Group all results by the dataset id in the filename
- Pick the ASDF file to use with the following priority: (`_metadata`, `_user_tools`, `<nothing>`)
- Throw a warning to the user if any ASDF files with older suffixes are found

Cadair:
fixes #496 

I am not super happy with the parsing code, so if anyone has any suggestions for making it better I am very open to improving it.


dkist/dataset/loader.py
```diff
@@ -14,6 +19,9 @@
     from asdf import ValidationError
 
 
+ASDF_FILENAME_PATTERN = r"^(?P<instrument>[A-Z-]+)_L1_(?P<timestamp>\d{8}T\d{6})_(?P<datasetid>[A-Z]+)(?P<suffix>_user_tools|_metadata)?.asdf$"
```
SolarDrew, line 22: Is the second `-` in the instrument group deliberate?

eigenbrot, line 22: Do our dataset ID's not have a fixed length?

dkist/dataset/loader.py
```diff
@@ -138,21 +146,91 @@ def _load_from_path(path: Path):
...
+                asdfs_to_load.append(base_path / f"{prefix}.asdf")
+            else:
+                # This branch should never be hit because the regex enumerates the suffixes
+                raise ValueError("Unknown suffix encountered.")  # pragma: no cover
+
+    # Throw a warning if we have skipped any files
+    if ignored_files := set(asdf_files).difference(asdfs_to_load):
+        warnings.warn(
+            f"ASDF files with old names ({', '.join([a.name for a in ignored_files])}) "
+            "were found in this directory and ignored.",
```
SolarDrew, line 225: Perhaps add a sentence here to advise users to delete the old files? Otherwise they'll keep on getting the warning, which I'm sure will annoy a non-zero number of people.

dkist/dataset/tests/test_load_dataset.py
```diff
@@ -114,3 +118,87 @@ def test_not_dkist_asdf(tmp_path):
...
+    ("wibble.asdf", False),
+    ])
+def test_asdf_regex(filename, match):
+    m = re.match(ASDF_FILENAME_PATTERN, filename)
+    assert bool(m) is match
+
+
+@pytest.mark.parametrize(("filenames", "indices"), [
+    # param[0] is list of filenames
+    # parram[1] is the indices in that list that should be used
```
SolarDrew, line 151:
```suggestion
    # param[1] is the indices in that list that should be used
```

dkist/dataset/loader.py
```diff
@@ -138,21 +146,91 @@ def _load_from_path(path: Path):
...
+            continue
+
+        # All the matches have to be checked
+        candidates.append(filepath)
+
+    # If we only have one match load it
+    if len(candidates) == 1:
+        asdfs_to_load += candidates
+    else:
+        # Now we group by prefix
```
eigenbrot, line 202: Small thing, but I _think_ "prefix" in the follow context just means "everything that isn't the suffix", which confused me a bit. Would `stem` be a better name (taking a hint from `pathlib.Path`)?

eigenbrot (approved):
IMO, what the parsing lacks in concision it makes up for in clarity (the good comments help, too!).

dkist/dataset/loader.py
Cadair: stem also feels overloaded because it's everything in the filename which isn't `.asdf`.

dkist/dataset/loader.py
Cadair: I think they have a minimum length, but if we mint enough datasets they will get longer iirc

SolarDrew (approved):
LGTM :+1:

dkist/dataset/loader.py
Cadair: yes, it's a literal `-` so it's matching anything from `A-Z` or `-`.

dkist/dataset/loader.py
eigenbrot: Yeah, good point. I'm fine with whatever.

dkist/dataset/loader.py
Cadair: Maybe @fraserwatson can remember what the minimum length is we could add that as a constraint.

fraserwatson: You are correct, the dataset IDs are at least 5 characters but there is no upper bound on the length in theory.

Hyzerputt (approved):

fraserwatson (approved):

dkist/dataset/loader.py
Cadair: I can not think of a good word.

PR #504, opened 2025-01-22
Correct `TiledDataset.plot` mosaic ordering

eigenbrot:
Currently `TiledDataset.plot` flattens the tile list and places them in a subplot based on matplotlib's ordering of the `add_subplot` index ordering. The result is a grid that starts in the upper left. Furthermore, the fix in [this inventory PR](https://bitbucket.org/dkistdc/dkist-inventory/pull-requests/61/) will swap the MINDEX keys so that the current `TiledDataset.plot` method would plot a row-major grid, which is counter to the column-major intention of the MINDEX keys.

This PR fixes the first issue and prepares for the update mentioned in the second by using `GridSpec` to be more explicit about how the subplot grid should be laid our and exactly which tiles go into each subplot. The result is a (MAXIS1, MAXIS2)-shaped grid (FITS ordering) where the origin is in the bottom left and the subplot at grid position (x, y) corresponds to the tile with MINDEX1=x, MINDEX2=y.

As an additional refinement, there is also a new option (`limits_from_wcs`) to swap axis limits so that the minimum values of the WCS axes are also in the bottom left. This was written to help display some DLNIRSP data where the at-rest data array is oriented strangely, but it should be applicable to all instruments.

To show some before and after images I've taken headers from a real DLNIRSP dataset and modified the data arrays to simply show the (MINDEX1, MINDEX2) value of each tile. Here is `.plot()` as it is now:

dkist_DL_wrong

Note that the labels are upside down and backwards and the tiles are not organized in an intuitive grid.

Here are the same data after the fix for how the subplot grid is constructed

dkist_DL_right_raw_lims

Note that the tiles are now in the right places, but the labels are still flipped around. This is because the axes are "backward", in a WCS sense.

Finally here is the same plot with `limits_from_wcs=True`

dkist_DL_right

The labels are now readable because the smallest WCS values are now at the bottom left of each subplot.


dkist/dataset/tiled_dataset.py
```diff
@@ -184,6 +206,9 @@ def plot(self, slice_index, share_zscale=False, fig=None, **kwargs):
         fig : `matplotlib.figure.Figure`
             A figure to use for the plot. If not specified the current pyplot
             figure will be used, or a new one created.
+        limits_from_wcs : `bool`
```
eigenbrot, line 209: I'm not convinced `limits_from_wcs` is the best name for this option. Suggestions welcome!

SolarDrew: Perhaps `auto-rotate` or similar?

SolarDrew (commented):
Thanks for this, @eigenbrot , it looks like a great fix. It needs some tests but no other comments from me :+1:

dkist/dataset/tiled_dataset.py
eigenbrot: Maybe `auto_orient`? I like where you're going, but I'm worried "rotate" might imply that it can rotate by something other than 180 degrees.

dkist/dataset/tiled_dataset.py
```diff
@@ -167,7 +170,26 @@ def _get_axislabels(ax):
                 ylabel = coord.get_axislabel() or coord._get_default_axislabel()
         return (xlabel, ylabel)
 
-    def plot(self, slice_index, share_zscale=False, fig=None, **kwargs):
+    @staticmethod
+    def _ensure_wcs_ordered_axis_lims(wcs: HighLevelWCSWrapper, ax: Axes) -> None:
```
Cadair, line 174: I am generally uneasy about this whole thing. `WCSAxes` is weird and I worry this is poking a bear.

Cadair, line 174:
```suggestion
    def _ensure_wcs_ordered_axis_lims(ax: WCSAxes) -> None:
```

dkist/dataset/tiled_dataset.py
```diff
@@ -167,7 +170,26 @@ def _get_axislabels(ax):
...
+        Adjust axis limits so the WCS values go from smaller numbers in the bottom left to higher numbers in the upper right.
+        """
+        xmin, xmax = ax.get_xlim()
+        ymin, ymax = ax.get_ylim()
+
+        world_xmin, world_ymin = wcs.low_level_wcs.pixel_to_world_values(xmin, ymin)
+        world_xmax, world_ymax = wcs.low_level_wcs.pixel_to_world_values(xmax, ymax)
+
+        new_xmin, new_ymin = wcs.low_level_wcs.world_to_pixel_values(min(world_xmin, world_xmax), min(world_ymin, world_ymax))
+        new_xmax, new_ymax = wcs.low_level_wcs.world_to_pixel_values(max(world_xmin, world_xmax), max(world_ymin, world_ymax))
```
Cadair, line 185:
```suggestion
        world_xmin, world_ymin = ax.wcs.pixel_to_world_values(xmin, ymin)
        world_xmax, world_ymax = ax.wcs.pixel_to_world_values(xmax, ymax)

        new_xmin, new_ymin = ax.wcs.world_to_pixel_values(min(world_xmin, world_xmax), min(world_ymin, world_ymax))
        new_xmax, new_ymax = ax.wcs.world_to_pixel_values(max(world_xmin, world_xmax), max(world_ymin, world_ymax))
```

dkist/dataset/tiled_dataset.py
```diff
@@ -167,7 +170,26 @@ def _get_axislabels(ax):
...
+
+        ax.set_xlim(new_xmin, new_xmax)
+        ax.set_ylim(new_ymin, new_ymax)
+
+        return
+
+    def plot(self, slice_index, share_zscale=False, fig=None, limits_from_wcs=True, **kwargs):
```
Cadair, line 192: I would like to throw a warning here if the user loads an old ASDF file with this new `.plot` to go and tell them to redownload a new ADSF.

eigenbrot: I'm fine with this idea, but I'm not sure how to implement it. A user will get wrong plots (and should thus be warned) iff these two conditions are BOTH true:

1. `dkist-processing-vbi` version is < 1.19.0
2. `dkist-inventory` version is < 1.?.?

We can check 1 easily, but I don't know about checking the `dkist-inventory` version. Ideas?

dkist/dataset/tiled_dataset.py
eigenbrot: Good idea!

Cadair (changes requested):
I am going to put this on hold until I can chat to @astrofrog about the best way to do the WCSAxes flipping.

dkist/dataset/tiled_dataset.py
```diff
@@ -167,7 +169,26 @@ def _get_axislabels(ax):
...
+        ymin, ymax = ax.get_ylim()
+
+        world_xmin, world_ymin = ax.wcs.pixel_to_world_values(xmin, ymin)
+        world_xmax, world_ymax = ax.wcs.pixel_to_world_values(xmax, ymax)
+
+        new_xmin, new_ymin = ax.wcs.world_to_pixel_values(min(world_xmin, world_xmax), min(world_ymin, world_ymax))
+        new_xmax, new_ymax = ax.wcs.world_to_pixel_values(max(world_xmin, world_xmax), max(world_ymin, world_ymax))
+
+        ax.set_xlim(new_xmin, new_xmax)
+        ax.set_ylim(new_ymin, new_ymax)
```
Cadair, line 187: Alright, after some deliberation:
```python
from astropy.wcs.utils import local_partial_pixel_derivatives
```
```suggestion
    local_derivs = local_partial_pixel_derivatives(ax.wcs, *[0]*ax.wcs.pixel_n_dim)
    
    _ = ax.get_xlabel()  # triggers a call to ax._update_tick_and_label_positions()
    
    for coord in ax.coords:
        if 'b' in coord.get_ticklabel_position():
            if local_derivs[coord.coord_index, 0] < 0:
                ax.invert_xaxis()
    
        if 'l' in coord.get_ticklabel_position():
            if local_derivs[coord.coord_index, 1] < 0:
                ax.invert_yaxis()

```

The main things this does differently is: that it does not rely on limits as inputs, which may be unreliable if the image is rotated; uses `invert_[xy]axis` rather than setting limits so the limits wont be affected by numerical error in the round trip transform; and uses `local_partial_pixel_derivatives` to calculate the direction of change of each coordinate.

dkist/dataset/tiled_dataset.py
Cadair: Given the changes I suggest to the algorithm above, I think something like `auto_invert_axes` or something?

dkist/dataset/tiled_dataset.py
eigenbrot: Very cool idea, but ran into some terrible nastiness trying to get it working for `astropy < 7.x`. Solution is just to expose an option to `.plot` to swap one or both of the axes. Let the user just say what they want.

changelog/504.feature.2.rst
Cadair, line 1: This now needs updating.

Cadair (changes requested):
This is now blocked on needing a warning if this changed plot ordering is used with currently available ASDF files.

Otherwise looks good.

Cadair (approved):

PR #507, opened 2025-01-28
Update dependancies and min Python

Cadair:


Cadair:
The required jobs will need updating in GH settings.

PR #513, opened 2025-01-30
Expose the ASDF history in meta

Cadair:


dkist/io/asdf/converters/tiled_dataset.py
```diff
@@ -5,16 +5,22 @@ class TiledDatasetConverter(Converter):
...
+        return TiledDataset(node["datasets"], node["inventory"], meta=node.get("meta"))
 
     def to_yaml_tree(cls, tiled_dataset, tag, ctx):
         tree = {}
-        tree["inventory"] = tiled_dataset._inventory
+        meta = copy.deepcopy(tiled_dataset.meta)
+        inventory = meta.pop("inventory")
+        # If the history key has been injected into the meta, do not save it
+        meta.pop("history")
+        tree["inventory"] = inventory
```
Cadair, line 23: If I am bumping the schema version here anyway, I might as well move inventory into meta.

dkist/io/asdf/converters/tiled_dataset.py
```diff
@@ -5,16 +5,22 @@ class TiledDatasetConverter(Converter):
     tags = [
         "tag:dkist.nso.edu:dkist/tiled_dataset-0.1.0",
         "asdf://dkist.nso.edu/tags/tiled_dataset-1.0.0",
+        "asdf://dkist.nso.edu/tags/tiled_dataset-1.1.0",
```
Cadair, line 8: In general with this schema bump I think we should standardise across Dataset and TiledDataset both in memory and on disk.

dkist/dataset/tiled_dataset.py
```diff
@@ -124,12 +126,19 @@ def flat(self):
         """
         return type(self)(self._data.flat, self.inventory)
 
+    @property
+    def meta(self):
```
SolarDrew, line 130: What does this get us compared to just having it be a normal attribute? Is it just so it's not writeable?

dkist/io/asdf/converters/dataset.py
```diff
@@ -63,6 +63,8 @@ def to_yaml_tree(self, dataset, tag, ctx):
                              "it was not constructed from a set of FITS files.")
         node = {}
         node["meta"] = dataset.meta or {}
+        # If the history key has been injected into the meta, do not save it
+        node["meta"].pop("history", None)
```
SolarDrew, line 67: Won't this mean the history is lost once someone writes a new asdf file? Or is that the point?

dkist/dataset/tiled_dataset.py
Cadair: Yep

dkist/io/asdf/converters/dataset.py
Cadair: Yeah. Especially given we don't have a write method at the moment.

SolarDrew: Fair point.

SolarDrew (approved):

Issue #265, opened 2023-06-06
Make the header table have more metadata


Cadair:
Things like units or `Time` columns should be possible, by processing the header spec in some way.

Things to consider:

* How to use the correct version of the spec, should this be written into the asdf?
* Do we want to make the header table more a more complex asdf object, which might hinder it ever being read by IDL or other languages?

PR #514, opened 2025-01-30
Asdf read speed

### LINKED ISSUE #500: Reading a DL-NIRSP ASDF is very slow
With a test file on my local system (with the profiler enabled) it took 600s, which is *insane* (it does take significantly less without the profiler).

Here are some excerpts from the profile (which is insanely large)
```
612.109 <module>  <ipython-input-10-8499de7ed805>:1
└─ 612.109 wrapper  functools.py:927
   └─ 612.109 _load_from_string  /home/stuart/Git/DKIST/dkist/dkist/dataset/loader.py:116
      └─ 612.109 _load_from_path  /home/stuart/Git/DKIST/dkist/dkist/dataset/loader.py:125
         ├─ 612.109 _load_from_asdf  /home/stuart/Git/DKIST/dkist/dkist/dataset/loader.py:158
         │  ├─ 611.866 open_asdf  asdf/_asdf.py:1622
         │  │  ├─ 611.861 AsdfFile._open_impl  asdf/_asdf.py:1006
         │  │  │  └─ 611.861 AsdfFile._open_asdf  asdf/_asdf.py:890
         │  │  │     ├─ 360.544 AsdfFile._validate  asdf/_asdf.py:670
```
```
         │  │  │     ├─ 114.634 tagged_tree_to_custom_tree  asdf/yamlutil.py:329
```
```
         │  │  │     ├─ 88.697 load_tree  asdf/yamlutil.py:373
```
```
         │  │  │     ├─ 39.880 find_references  asdf/reference.py:108
```
```
         │  │  │     ├─ 7.834 Manager.read  asdf/_block/manager.py:337
```

So a significant amount of time is in the validation of the file on read, followed by the conversion of the tree to high-level objects and a good chunk in parsing the yaml and finding all the references in the yaml.

The obvious win would be to disable validation on read, but we should think about the trade off more.

SolarDrew:
Fixes #500


Cadair:
I just did a quick experiment locally and if we convert the Table to a numpy recarray before we save it (`Table.as_array()`) then asdf will automatically only write the one binary block with the full data and will save slices in the tree as references

<details>
<summary>Full code</summary>

```python
import dkist; from dkist.data.sample import VBI_AJQWW
tds = dkist.load_dataset(VBI_AJQWW)

whole_table = tds.combined_headers
import asdf

small1 = whole_table[0:10]
small2 = whole_table[10:20]

new_tree = {"whole": whole_table, "small1":small1, "small2": small2}
with asdf.AsdfFile(tree=new_tree) as af:
    af.write_to("test.asdf")

<duplicates the data>

whole_array = whole_table.as_array()
array_tree = {"whole": whole_array, "small1":whole_array[0:10], "small2": whole_array[10:20]}
with asdf.AsdfFile(tree=array_tree) as af:
    af.write_to("array.asdf")

<does not duplicate the data>
```
</details>


For a small example:
```python
whole_table2 = whole_table[["INSTRUME", "DATE-AVG"]]
whole_array2 = whole_table2.as_array()
array_tree2 = {"whole": whole_array2, "small1":whole_array2[0:10], "small2": whole_array2[10:20]}
with asdf.AsdfFile(tree=array_tree2) as af:
    af.write_to("array2.asdf")
```
yields this asdf:

```
#ASDF 1.0.0
#ASDF_STANDARD 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
asdf_library: !core/software-1.0.0 {author: The ASDF Developers, homepage: 'http://github.com/asdf-format/asdf',
  name: asdf, version: 3.5.0}
history:
  extensions:
  - !core/extension_metadata-1.0.0
    extension_class: asdf.extension._manifest.ManifestExtension
    extension_uri: asdf://asdf-format.org/core/extensions/core-1.5.0
    manifest_software: !core/software-1.0.0 {name: asdf_standard, version: 1.1.1}
    software: !core/software-1.0.0 {name: asdf, version: 3.5.0}
small1: !core/ndarray-1.0.0
  source: 0
  datatype:
  - byteorder: little
    datatype: [ucs4, 3]
    name: INSTRUME
  - byteorder: little
    datatype: [ucs4, 26]
    name: DATE-AVG
  byteorder: big
  shape: [10]
small2: !core/ndarray-1.0.0
  source: 0
  datatype:
  - byteorder: little
    datatype: [ucs4, 3]
    name: INSTRUME
  - byteorder: little
    datatype: [ucs4, 26]
    name: DATE-AVG
  byteorder: big
  shape: [10]
  offset: 1160
whole: !core/ndarray-1.0.0
  source: 0
  datatype:
  - byteorder: little
    datatype: [ucs4, 3]
    name: INSTRUME
  - byteorder: little
    datatype: [ucs4, 26]
    name: DATE-AVG
  byteorder: big
  shape: [27]
...
[BINARY BLOCK]
%YAML 1.1
---
- 1231
...
```

notice the `source: 0` for all the arrays.


---

I think this might be a good idea.

My main worry is that it severely limits how rich we can make the metadata table, i.e. #265 becomes something custom we have to glue on the side rather than being able to use built-in features of astropy Table.

However I think this approach has many advantages:

* Obvious performance improvements for single tables (probably), but especially for what you are doing on this PR.
* We can still convert up to a table either in the converter or in Dataset itself, without copying the memory.
* This is almost certainly more portable to other languages, as the ndarray tag and schema are in the core spec. It would be worth testing to see what happens in IDL.

dkist/io/asdf/converters/tiled_dataset.py
```diff
@@ -1,20 +1,33 @@
...
+        else:
+            headers = vstack([Table(ds.headers) for ds in row for row in node["datasets"]])
+
+        return TiledDataset(node["datasets"], node["inventory"], headers)
 
     def to_yaml_tree(cls, tiled_dataset, tag, ctx):
         tree = {}
         tree["inventory"] = tiled_dataset._inventory
         tree["datasets"] = tiled_dataset._data.tolist()
+        tree["headers"] = tiled_dataset.combined_headers.as_array()
```
Cadair, line 32: We should do this for dataset too.

dkist/io/asdf/resources/schemas/dataset-1.2.0.yaml
```diff
@@ -0,0 +1,50 @@
...
+  meta:
+    description: Dataset metadata, describing the whole dataset.
+    type: object
+    properties:
+      headers:
+        description: A table of all the headers for the constituent files.
+        anyOf:
+          - tag: "tag:astropy.org:astropy/table/table-1.*"
+          - tag: 'tag:stsci.edu:asdf/core/ndarray-1.*'
+          - null: true
```
Cadair, line 35: I think this isn't where we want to end up, we want to have this always be it's own recarray (either a whole one or a view into another one)

dkist/dataset/tiled_dataset.py
```diff
@@ -219,14 +213,6 @@ def inventory(self) -> dict[str, Any]:
         """
         return self._meta["inventory"]
 
-    @property
-    def combined_headers(self) -> Table:
-        """
-        A single `astropy.table.Table` containing all the FITS headers for all
-        files in this dataset.
-        """
-        return vstack([ds.meta["headers"] for ds in self._data.compressed()])
```
Cadair, line 228: put this back but it should be:

```python
if self.meta.get("headers") is None:
    <vstack...>
return Table(self.meta["headers"], copy=False)
```

dkist/io/asdf/resources/schemas/dataset-1.2.0.yaml
Cadair, line 35: Given where we are ending up with this, this should only ever be a structured array.
```suggestion
          - tag: 'tag:stsci.edu:asdf/core/ndarray-1.*'
```

dkist/io/asdf/resources/schemas/tiled_dataset-1.3.0.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+      inventory:
+        description: A copy of the inventory record for this dataset.
+        type: object
+
+      headers:
+        description: A table of all the headers for the constituent files.
+        tag: 'tag:stsci.edu:asdf/core/ndarray-1.*'
+
+    required: [inventory]
```
Cadair, line 39: I think we should require headers here too

dkist/dataset/dataset.py
```diff
@@ -128,6 +128,8 @@ def __init__(self, data, wcs=None, uncertainty=None, mask=None, meta=None,
         if "inventory" not in meta:
             raise ValueError("The meta dict must contain the inventory record.")
 
+        self._is_mosaic_tile = is_tile
+
```
Cadair, line 132:
```suggestion
```

dkist/dataset/dataset.py
```diff
@@ -105,7 +105,7 @@ class Dataset(NDCube):
     _file_manager = FileManagerDescriptor(default_type=DKISTFileManager)
 
     def __init__(self, data, wcs=None, uncertainty=None, mask=None, meta=None,
-                 unit=None, copy=False):
+                 unit=None, copy=False, is_tile=False):
```
Cadair, line 108: Kill this off?
```suggestion
                 unit=None, copy=False):
```

dkist/dataset/tiled_dataset.py
```diff
@@ -150,6 +128,27 @@ def __init__(
...
+                [Table(ds.headers) if ds else Table({}) for ds in self._data.compressed()]
+            )
+        # Then distribute headers (back) out to component Datasets as slices of the main Table
+        # for row in dataset_array:
+        #     for ds in row:
+
+        offset = 0
+        for ds in self._data.compressed():
+            if ds:
+                if isinstance(ds.headers, dict) and "offset" in ds.headers.keys():
```
Cadair, line 145: I don't think this should be here, any offset / shape thing should just be in the asdf / asdf converters.

dkist/dataset/tiled_dataset.py
```diff
@@ -150,6 +128,27 @@ def __init__(
...
+        # Otherwise stack the headers saved for component Datasets
+        if meta.get("headers") is None:
+            meta["headers"] = vstack(
+                [Table(ds.headers) if ds else Table({}) for ds in self._data.compressed()]
+            )
+        # Then distribute headers (back) out to component Datasets as slices of the main Table
+        # for row in dataset_array:
+        #     for ds in row:
+
+        offset = 0
```
Cadair, line 142: I think we should only be doing this split out if `meta.get("headers") is None`, i.e if we haven't loaded from asdf.

dkist/dataset/tiled_dataset.py
SolarDrew: Fairly sure you said it should be in the constructor but I don't remember why :smile:

dkist/dataset/tiled_dataset.py
SolarDrew: If we don't then the tile headers and the combined_headers aren't sharing memory because vstack creates a copy.

Cadair: what I mean is we should only do this next bit if we have vstacked not if we already had combined headers (in which case we should assume they are correct).

dkist/io/asdf/resources/schemas/dataset-1.2.0.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+  meta:
+    description: Dataset metadata, describing the whole dataset.
+    type: object
+    properties:
+      headers:
+        description: A table of all the headers for the constituent files.
+        anyOf:
+          - tag: 'tag:stsci.edu:asdf/core/ndarray-1.*'
+          - null: true
```
Cadair, line 34:
```suggestion
          - type: object
            required: [offset, size]
            addtionalProperties: false
```

dkist/dataset/tiled_dataset.py
```diff
@@ -150,6 +128,25 @@ def __init__(
...
+            # Then distribute headers (back) out to component Datasets as slices of the main Table
+            offset = 0
+            for ds in self._data.compressed():
+                if ds:
+                    if isinstance(ds.headers, dict) and "offset" in ds.headers.keys():
+                        thisoffset = ds.meta["headers"]["offset"]
+                        s = ds.meta["headers"]["size"]
+                        ds.meta["headers"] = meta["headers"][thisoffset:thisoffset+s]
+                    else:
+                        ds.meta["headers"] = meta["headers"][offset:offset+len(ds.files or [])]
```
Cadair, line 148:
```suggestion
                    ds.meta["headers"] = meta["headers"][offset:offset+len(ds.files or [])]
```

dkist/io/asdf/converters/tiled_dataset.py
```diff
@@ -17,6 +18,12 @@ def from_yaml_tree(cls, node, tag, ctx):
 
         # Support old files without meta, but with inventory
         meta = node.get("meta", {})
+
+        for row in node["datasets"]:
+            for ds in row:
+                if ds:
+                    ds._is_mosaic_tile = True
+
```
Cadair, line 26:
```suggestion
```

dkist/dataset/tiled_dataset.py
```diff
@@ -150,6 +128,18 @@ def __init__(
         self._data = np.ma.masked_array(dataset_array, dtype=object, mask=mask)
         meta = meta or {}
         inventory = meta.get("inventory", inventory or {})
+
+        # If headers are saved as one Table for the whole TiledDataset, use those first
+        # Otherwise stack the headers saved for component Datasets
+        if meta.get("headers") is None:
+            ds_headers = [Table(ds.headers) if ds else Table({}) for ds in self._data.compressed()]
```
Cadair, line 135: Should we be using `if ds` here?

When is ds falsey? isn't .compressed() removing anywhere where the mask is True? So why would we have mask False and not a valid ds?

dkist/dataset/tiled_dataset.py
```diff
@@ -185,6 +175,10 @@ def _validate_component_datasets(datasets, inventory):
                 raise ValueError("The inventory records of all the datasets do not match the one passed to TiledDataset")
         return True
 
+    @property
+    def combined_headers(self):
```
Cadair, line 179: This has lost it's docstring and type hint.

dkist/io/asdf/converters/tiled_dataset.py
```diff
@@ -30,6 +32,13 @@ def to_yaml_tree(cls, tiled_dataset, tag, ctx):
         # If the history key has been injected into the meta, do not save it
         meta.pop("history", None)
         tree["meta"] = meta
+        tree["meta"]["headers"] = tiled_dataset.combined_headers
```
Cadair, line 35: We don't need this right? `combined_headers` is just `return meta["headers"]`?
```suggestion
```

dkist/io/asdf/resources/schemas/dataset-1.2.0.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+  meta:
+    description: Dataset metadata, describing the whole dataset.
+    type: object
+    properties:
+      headers:
+        description: A table of all the headers for the constituent files.
+        anyOf:
+          - tag: "tag:astropy.org:astropy/table/table-1.*"
+          - tag: 'tag:stsci.edu:asdf/core/ndarray-1.*'
```
Cadair, line 34: We never went this way right?
```suggestion
```

dkist/dataset/tiled_dataset.py
```diff
@@ -150,6 +128,18 @@ def __init__(
         self._data = np.ma.masked_array(dataset_array, dtype=object, mask=mask)
         meta = meta or {}
         inventory = meta.get("inventory", inventory or {})
+
+        # If headers are saved as one Table for the whole TiledDataset, use those first
+        # Otherwise stack the headers saved for component Datasets
+        if meta.get("headers") is None:
```
Cadair, line 134: I think this bit is missing an explicit test? It would be good to add one which makes sure that the dkist-inventory workflow behaves as expected.

dkist/io/asdf/resources/manifests/dkist-1.5.0.yaml
Cadair, line 1: Also, we should add new test data for new dataset and tiled_dataset schemas so we have tests which test all old versions of the schemas.

Cadair: and apparently for some historical ones as well :see_no_evil:

dkist/dataset/tiled_dataset.py
```diff
@@ -181,10 +171,18 @@ def _validate_component_datasets(datasets, inventory):
         for ds in datasets[1:]:
             if ds.wcs.world_axis_physical_types != pt_1:
                 raise ValueError("The physical types do not match between all datasets")
-            if ds.meta["inventory"] and ds.meta["inventory"] is not inventory:
+            if ds.meta["inventory"] and ds.meta["inventory"] != inventory:
```
Cadair, line 174: I'd rather not make this change. The "is not" was deliberate, it should be the exact same dict in all places as it's the same record.

dkist/io/asdf/converters/tiled_dataset.py
```diff
@@ -30,6 +32,12 @@ def to_yaml_tree(cls, tiled_dataset, tag, ctx):
         # If the history key has been injected into the meta, do not save it
         meta.pop("history", None)
         tree["meta"] = meta
+        # Go into dataset header attributes and replace with {"offset": ..., "size": ...}
```
Cadair, line 35: heh, I just realised doing this will affect what's in memory:
```python
ds = TiledDataset(...)
asdf.write(ds)
ds[0,0].headers == {"offset": ...}
```
which isn't ideal.

I think this might mean it's custom tag for us.

Cadair (changes requested):
Closer than we've ever been.

dkist/dataset/tiled_dataset.py
SolarDrew: Does it matter if the content is the same?

Cadair: it matters if you save it back to asdf as it wont dedupe. Also it's not as technically correct.

dkist/io/asdf/converters/tiled_dataset.py
```diff
@@ -33,7 +33,7 @@ def to_yaml_tree(cls, tiled_dataset, tag, ctx):
         meta.pop("history", None)
         tree["meta"] = meta
         # Copy the data as well so we aren't editing dataset headers in place
-        datasets = copy.copy(tiled_dataset._data)
+        datasets = copy.deepcopy(tiled_dataset._data)
```
Cadair, line 36: oh really? This makes me worry it's going to copy more stuff than I expected and take more memory.

SolarDrew: Don't disagree but just `copy` doesn't copy enough stuff it seems :shrug:

Cadair: What doesn't it copy?!

Cadair: oh sure, copy would only copy the array and not the references to the dataset's within.

We want the copy to copy the dataset objects but not the stuff contained within them :thinking:

SolarDrew: Indeed. Corresponding commits incoming.

SolarDrew:
pre-commit.ci autofix

SolarDrew:
If I remember rightly, the current notebooks failure is a known sunpy issue and not a problem with this PR, which means this is actually finally ready to go, pending final review I guess.

dkist/dataset/tiled_dataset.py
```diff
@@ -150,6 +132,23 @@ def __init__(
...
+            offsets = np.cumsum([0] + sizes[:-1])
+            meta["headers"] = vstack(ds_headers)
+
+            # Then distribute headers (back) out to component Datasets as slices of the main Table
+            for i, ds in enumerate(self._data.compressed()):
+                ds.meta["headers"] = meta["headers"][offsets[i]:offsets[i]+sizes[i]]
+        elif isinstance(self._data.compressed()[0].headers, dict):
+            for ds in self._data.compressed():
+                offset, size = ds.headers["offset"], ds.headers["size"]
+                ds.meta["headers"] = meta["headers"][offset:offset+size]
```
Cadair, line 150: This should move into the converter.

dkist/dataset/tiled_dataset.py
```diff
@@ -75,7 +76,10 @@ def __getitem__(self, slice_):
                 continue
             new_data.flat[i] = ds[slice_]
 
-        return TiledDataset(new_data, meta=self.meta, mask=self.data.mask)
+        meta = copy.copy(self.meta)
+        meta["headers"] = None
```
Cadair, line 80: Why?

tox.ini
```diff
@@ -54,6 +54,8 @@ deps =
     figure-!devdeps: mpl-animators==1.1.1
     figure-!devdeps: astropy==6.1.1
     figure-!devdeps: ndcube==2.2.2
+	# Patch in ndcube 2.3.2 needed for one of the botebook plots to run in a sensible time
+	build_docs-notebooks: ndcube==2.3.2
```
Cadair, line 58:
```suggestion
    # Patch in ndcube 2.3.2 needed for one of the botebook plots to run in a sensible time
    build_docs-notebooks: ndcube>=2.3.2
```

Cadair (changes requested):

Cadair:
I think this is now ready, we just need to coordinate releasing it with the data center team.

eigenbrot (approved):
I'm approving just based on the fact that it works on real DL data.

SolarDrew:
build_docs failure is fixed in #574 . Not sure what's up with the oldestdeps one.

SolarDrew:
> Merging https://github.com/DKISTDC/dkist/pull/514 will degrade performances by 10.98%

WWWWWHHHHHHHYYYY!?!?!?

SolarDrew:
No, that'll do it

PR #518, opened 2025-02-05
Add a script for regenerating and uploading sample data

Cadair:


tools/update_sample_data.py
```diff
@@ -0,0 +1,102 @@
...
+
+    for did, props in datasets.items():
+        res = Fido.search(a.dkist.Dataset(did))
+        asdf_file = Fido.fetch(res, path=working_directory / "{dataset_id}", progress=False, overwrite=False)
+
+        ds = dkist.load_dataset(asdf_file)
+        if "slice" in props:
+            ds = ds[props["slice"]]
+        if "tile_slice" in props:
+            ds = ds.slice_tiles[props["tile_slice"]]
```
SolarDrew, line 54: Do we need separate variables here? We could just do the indexing in the if tiled/else block.

SolarDrew (approved):
Minor structural comment but otherwise looks good :+1:

tools/update_sample_data.py
Cadair: Well you could also slice down how many tiles you have like this with `slice`.

Cadair: Yeah, the idea here is to allow this to scale to other future datasets we might want to include, for example if we have a DL one we might only want to ship a subset of the tiles for a moasic.

tools/update_sample_data.py
```diff
@@ -0,0 +1,102 @@
...
+    working_directory.mkdir(parents=True, exist_ok=True)
+    sample_files_for_upload = []
+
+    for did, props in datasets.items():
+        res = Fido.search(a.dkist.Dataset(did))
+        asdf_file = Fido.fetch(res, path=working_directory / "{dataset_id}", progress=False, overwrite=False)
+
+        ds = dkist.load_dataset(asdf_file)
+        if "slice" in props:
+            ds = ds[props["slice"]]
```
eigenbrot, line 52: Don't we need to save the sliced dataset metadata file? I.e., right now the metadata file provided as part of the sample data would refer to more than the sample data files, right?

Cadair: That's correct, but that's ok, the idea is that some files are missing which is a supported usecase.

PR #519, opened 2025-02-05
Bump gwcs version to 0.24

Cadair:


braingram:
https://github.com/DKISTDC/dkist/blob/5b0cd54e5e150307d94a32c43c1b3dd1e5cd7d3b/dkist/dataset/loader.py#L25-L28
could be removed with the asdf bump.

Cadair:
Thanks @braingram kinda got cornered into a larger set of bumps than I was planning for.

docs/whatsnew/1.10.rst
eigenbrot, line 1: Should "ASDF file" be replaced with "metadata file" in this doc? I see in some places it's "metadata ASDF", which I'm fine with, but that should be used everywhere.

docs/whatsnew/1.10.rst
```diff
@@ -0,0 +1,43 @@
...
+ASDF File Updates
+=================
+
+This release of ``dkist`` includes compatibility with new metadata ASDF files generated after 2025/02/07.
+
+.. warning::
+
+   You should re-download ASDF files when upgrading to this version of ``dkist``, particularly for VBI data.
+
+These ASDF files change the ordering of the tiles inside a `.TiledDataset` object to correctly account for the differences between column-major (in FITS files) and row-major ordering (in numpy), this release of ``dkist`` updates the `.TiledDataset.plot` method to correctly plot these tiles in the expected order.
```
eigenbrot, line 16: I think a semicolon would be better in "...(in numpy); this release..."

dkist/dataset/tiled_dataset.py
```diff
@@ -195,7 +196,7 @@ def plot(self, slice_index, share_zscale=False, figure=None, swap_tile_limits: L
         slice_index : `int`, sequence of `int`s or `numpy.s_`
             Object representing a slice which will reduce each component dataset
             of the TiledDataset to a 2D image. This is passed to
-            ``TiledDataset.slice_tiles``
+            ``TiledDataset.slice_tiles``, if each tile is already 2D pass ``slice_index=...`.
```
eigenbrot, line 199: Why `...` instead of `None`?

Cadair: Because `None` in numpy adds a dummy dimension, where as `...` is "the rest of the dimensions".

eigenbrot: Ah, gotcha. Thanks

docs/whatsnew/1.10.rst
Cadair: I think it reads correctly, I don't want to be overzealous with it.

eigenbrot (approved):

PR #523, opened 2025-02-08
Fixes related to sample data updates

### LINKED ISSUE #522: Add a way to clear the sample data cache
This is currently an issue because we have updated it.

I added [this](https://github.com/DKISTDC/dkist/pull/523/files#diff-5e33a88fa709d4f49d321e02c1f0d4f0483fba6ff2549a947d68e2d1c799d046R23) as the bare minimalist possible approach.

I would like a way for us to be able to push an update to the user tools and have the sample data cache be cleared for everyone. The best way I have thought to do this is have a date string stored in the `_sample.py` somewhere and if the write time on the .tar files is older than that date then we force a redownload.

### LINKED ISSUE #527: Update Figure test hashes (they use the new sample data) and remove `pytest.warns`

Cadair:
Closes #522 
Closes #527

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -16,7 +16,6 @@ jobs:
         with:
           python-version: '3.12'
       - run: python -m pip install .[tests] pytest-codspeed 'numpy<2'
-      - run: python -c "from dkist.data.sample import download_all_sample_data; download_all_sample_data()"
       - name: Run benchmarks
         uses: CodspeedHQ/action@v3
         with:
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -20,4 +20,4 @@ jobs:
         uses: CodspeedHQ/action@v3
         with:
           token: ${{ secrets.CODSPEED_TOKEN }}
-          run: "pytest -vvv -r fEs --pyargs dkist --codspeed"
+          run: "pytest -vvv -r fEs --pyargs dkist --codspeed --remote-data=any"
--- a/dkist/tests/test_benchmarks.py
+++ b/dkist/tests/test_benchmarks.py
@@ -41,6 +41,7 @@ def test_plot_dataset(benchmark, axes, visp_dataset_no_headers, tmp_path):
 
 
 @pytest.mark.benchmark
+@pytest.mark.remote_data
 def test_dataset_compute_data_full_files(benchmark):
     """
     Note that although this will load all the files to compute the data, the
@@ -48,7 +49,8 @@ def test_dataset_compute_data_full_files(benchmark):
     because it doesn't support that. This test therefore only assesses the
     performance of the compute step.
     """
-    from dkist.data.sample import VISP_BKPLX
+    from dkist.data.sample import VISP_BKPLX, download_all_sample_data
+    download_all_sample_data()
     ds = load_dataset(VISP_BKPLX)[0, :15]
     benchmark(ds.data.compute)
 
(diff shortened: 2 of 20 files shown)


Cadair:
pre-commit.ci autofix

pyproject.toml
```diff
@@ -24,7 +24,7 @@ dependencies = [
   # files with them unpinned so they are now required to read generated asdf
   # files.
   "asdf>=3.3.0",  # Required by gwcs 0.24
-  "asdf-astropy>=0.5.0",  # Required by gwcs 0.24
+  "asdf-astropy>=0.5.0,!=0.7.0",  # Required by gwcs 0.24
```
Cadair, line 27: This apparently didn't fix the problem:
```suggestion
  "asdf-astropy>=0.5.0",  # Required by gwcs 0.24
```

SolarDrew:
Note to self for later: I've marked the benchmarks that need sample data with `remote_data` but then I'm not sure when they're actually going to run, so I need to look into that.

dkist/tests/test_benchmarks.py
```diff
@@ -41,23 +41,27 @@ def plot_and_save_fig(ds=visp_dataset_no_headers, axes=axes):
...
 def test_dataset_compute_data_full_files(benchmark):
     """
     Note that although this will load all the files to compute the data, the
     file IO overhead is *not* included in codspeed's timing of the benchmark,
     because it doesn't support that. This test therefore only assesses the
     performance of the compute step.
     """
-    from dkist.data.sample import VISP_BKPLX
+    from dkist.data.sample import VISP_BKPLX, download_all_sample_data
+    download_all_sample_data()
```
Cadair, line 53: We don't need to explicitly call this, the import will do it.

.github/workflows/codspeed.yml
```diff
@@ -16,7 +16,6 @@ jobs:
         with:
           python-version: '3.12'
       - run: python -m pip install .[tests] pytest-codspeed 'numpy<2'
-      - run: python -c "from dkist.data.sample import download_all_sample_data; download_all_sample_data()"
```
Cadair, line 19: I don't think we need to remove this. This means that the data get's downloaded in it's own step and then the rest of the benchmarks run can be offline.

tox.ini
```diff
@@ -61,7 +61,6 @@ commands_pre =
     oldestdeps: minimum_dependencies dkist --filename requirements-min.txt
     # We need to pin down the cryptography transitive dependency because of globus
     oldestdeps: pip install -r requirements-min.txt cryptography<42
-    figure: python -c "from dkist.data.sample import download_all_sample_data; download_all_sample_data()"
```
Cadair, line 64: Not sure we need to pull this either?

.github/workflows/codspeed.yml
SolarDrew: I thought this was what was causing the problems though? If we bring this back then we probably don't need the remote_data mark on at least some of the ones I've added it to, because it won't need to download the data.

Cadair: It should have the remote data mark, because it wont work without the sample data (i.e if someone runs it locally). But there's no reason why we can't also pre-download the sample data in a separate step of the workflow.

Cadair: Also to make the benchmarks run with remote data, add `--remote-data=any` to the pytest line at the end of this file.

pytest.ini
```diff
@@ -70,3 +70,5 @@ filterwarnings =
     ignore:pkg_resources is deprecated as an API.*:DeprecationWarning
     ignore:Deprecated call to .*pkg_resources\.declare_namespace.*mpl_toolkits.*:DeprecationWarning
     ignore:Extension .*sunpy-1\.0\.0:asdf.exceptions.AsdfManifestURIMismatchWarning
+    # This warning should go away in asdf-astropy 0.7.1 and we can stop being ignored then
+    ignore::asdf.exceptions.AsdfPackageVersionWarning
```
braingram, line 74: FYI: 0.7.1 with the fix is up on pypi. https://pypi.org/project/asdf-astropy/0.7.1/

Cadair, line 74:
```suggestion
```

SolarDrew (approved):

PR #567, opened 2025-05-15
Support for level 2 data

SolarDrew:


dkist/dataset/inversion.py
```diff
@@ -0,0 +1,201 @@
...
+
+        return textwrap.dedent(s).format(quants_repr, profiles_repr)
+
+    def __getitem__(self, aslice):
+        new_inv = super().__getitem__(aslice)
+        new_inv.profiles = self.profiles
+
+        return new_inv
+
+    def plot_profiles(
```
Cadair, line 89: Wouldn't this be better off on the `.profiles` property?

SolarDrew: Well that makes a lot of sense.

dkist/net/client.py
```diff
@@ -293,3 +293,60 @@ def register_values(cls):
         }
 
         return {**return_values, **get_search_attrs_values()}
+
+
+class DKISTLevel2Client(DKISTClient):
```
Cadair, line 298: I'd probably not include this at until there's an API.

dkist/io/asdf/converters/inversion.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def from_yaml_tree(self, node, tag, ctx):
+        from dkist.dataset.inversion import Inversion
+
+        # We are "promoting" the quantities NDCollection object to being the Inversion object
+        quantities = node["quantities"]
+        aligned_axes = tuple(quantities.aligned_axes[key] for key in quantities.keys())
+        return Inversion(
+            node["quantities"],
+            meta=node["meta"],
```
Cadair, line 16: We need to handle the `node["quantities"].meta` part of the tree here, probably by adding it to `node["meta"]`

dkist/dataset/loader.py
```diff
@@ -140,6 +140,8 @@ def _load_from_path(path: Path, *, ignore_version_mismatch=False):
     if not path.is_dir():
         if not path.exists():
             raise ValueError(f"{path} does not exist.")
+        if "_L2_" in path.name:
+            return _load_l2_from_asdf(Path(path), ignore_version_mismatch=ignore_version_mismatch)
```
Cadair: oh we need a better solution for this

pytest.ini
```diff
@@ -75,3 +75,4 @@ filterwarnings =
     ignore:Deprecated call to .*pkg_resources\.declare_namespace.*mpl_toolkits.*:DeprecationWarning
     ignore:Signature b:UserWarning
     ignore:Reshaping is producing a large chunk:dask.array.core.PerformanceWarning
+	ignore:The chararray class is deprecated:DeprecationWarning
```
Cadair:
```suggestion
	ignore:The chararray class is deprecated:DeprecationWarning:astropy.io.fits
```
?

dkist/io/asdf/resources/schemas/profiles-0.1.0.yaml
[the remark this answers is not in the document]
SolarDrew: In principle yes. Unfortunately there's a practical hurdle which is that the schema infrastructure tests (#631)  require the number of schemas to equal the number of converter classes, and we need to keep the latter. I doubt it's an insurmountable problem but I don't think it's worth fixing now, and having the schema in place gives us scope to deviate from ndcollection easily.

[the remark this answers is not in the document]
Cadair: Ah here we go, we can reference the upstream schema, while keeping this file, which lets us extend it in the future if we want.

Like this:

https://github.com/asdf-format/asdf-coordinates-schemas/blob/3ad7f68a3a7ad3f688820d4b6c1b1cd3c46eebb7/resources/schemas/frames/cirs-1.1.0.yaml#L35-L36

So in this case I think it would be:
```yaml
$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
id: "asdf://dkist.nso.edu/schemas/profiles-0.1.0"
title: Represents a DKIST Level 2 inversion's source profiles

$ref: "asdf://sunpy.org/ndcube/schemas/ndcollection-1.0.0"
```

and that would be it.

PR #568, opened 2025-05-22
Update transform schemas to support transform-1.4.0 schema

braingram:
A new version of asdf-transform-schemas will soon be released with a new transform-1.4.0 schema. See https://github.com/asdf-format/asdf-transform-schemas/pull/120 for more details on the motivation for the new schema (tldr; this decouples the transform schemas from the core schemas in the standard to allow easier fixes for things like the ndarray mask bug: https://github.com/asdf-format/asdf/issues/1909).

This PR updates the dkist wcs schemas to use the new transform-1.4.0 schema.

A related asdf-astropy PR (https://github.com/astropy/asdf-astropy/pull/279) includes checks to verify the updates don't break the current dkist release due to a sort of self-referencing issue with the now old transform-1.3.0 schema. That PR makes the assumption that the dkist changes in this PR will end up in manifest 1.5.0.

This PR is part of an effort to untangle some dependencies in transform schemas. See https://github.com/asdf-format/asdf-wcs-schemas/pull/68 for probably too much context.

### DIFF
--- a/changelog/568.feature.rst
+++ b/changelog/568.feature.rst
@@ -1,0 +1,1 @@
+Add support for new transform schemas in asdf-transform-schemas 0.6.0.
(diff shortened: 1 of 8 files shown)


Cadair (approved):
lgtm, though the bump to asdf-transform-schemas means we need to update the asdf-maker service in prod.

PR #574, opened 2025-07-03
FileManager repr

SolarDrew:
Return of a basic FileManager repr. Plenty of room for improvement here, which can be in this PR or a later one.

### DIFF
--- a/dkist/io/file_manager.py
+++ b/dkist/io/file_manager.py
@@ -43,6 +43,12 @@ class DKISTFileManager:
     """
     __slots__ = ["_fm", "_inventory_cache", "_ndcube"]
 
+    def __str__(self) -> str:
+        return f"FileManager containing {len(self)} files with shape {self.shape}"
+
+    def __repr__(self) -> str:
+        return f"{object.__repr__(self)}\n{self}"
+
     @classmethod
     def from_parts(cls, fileuris, target, dtype, shape, *, loader, basepath=None, chunksize=None):
         return cls(
@@ -82,7 +88,6 @@ class DKISTFileManager:
             "__eq__",
             "__len__",
             "__str__",
-            "__repr__",
             "fileuri_array",
             "shape",
             "output_shape",
--- a/dkist/io/file_manager.py
+++ b/dkist/io/file_manager.py
@@ -44,7 +44,7 @@ class DKISTFileManager:
     __slots__ = ["_fm", "_inventory_cache", "_ndcube"]
 
     def __str__(self) -> str:
-        return f"FileManager containing {len(self)} files with shape {self.shape}"
+        return f"DKISTFileManager containing {len(self)} files with each array having shape {self.shape}"
 
     def __repr__(self) -> str:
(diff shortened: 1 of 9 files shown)


dkist/io/file_manager.py
```diff
@@ -43,6 +43,12 @@ class DKISTFileManager:
     """
     __slots__ = ["_fm", "_inventory_cache", "_ndcube"]
 
+    def __str__(self) -> str:
+        return f"DKISTFileManager containing {len(self)} files with each array having shape {self.shape}"
```
Cadair, line 47: I'd definitely put basepath in this as it's the thing the user is most likely to change / care about.

I'm not sure if the array shape or the dimensionality of the files array is the most interesting thing either. 🤔

tox.ini
```diff
@@ -54,6 +54,7 @@ deps =
     figure-!devdeps: mpl-animators==1.1.1
     figure-!devdeps: astropy==6.1.1
     figure-!devdeps: ndcube==2.2.2
+    build_docs-notebooks: git+https://github.com/sunpy/ndcube
```
Cadair, line 57: why?

dkist/io/file_manager.py
Cadair: you are phrasing this as "the array in each FITS file is this shape" is that what it is or is it "the array of filenames is this shape"?

tox.ini
SolarDrew: Because without recent additions to ndcube one of the plots in the visualisation notebook takes too long and times out, which breaks the build. I thought we'd fixed this for the workshop PR so I'm not sure why it's a problem again now. Obviously this isn't a good long-term solution.

Cadair: I don't think there's a fix to ndcube which is newer than the last bugfix release though? Let's not fold this in here, put it in it's own PR if we really need it?

dkist/io/file_manager.py
SolarDrew: The latter. I think both are useful pieces of information and to be honest I'm not sure what other information you're looking for if you're printing the filemanager, other than basepath.

SolarDrew: Oh actually it's the latter as it stands, but that's not what I intended it to be I think :thinking:

dkist/net/client.py
```diff
@@ -98,6 +98,11 @@ def _process_table(results: "DKISTQueryResponseTable") -> "DKISTQueryResponseTab
                 results[colname][none_values] = np.nan
             results[colname] = u.Quantity(results[colname], unit=unit)
 
+        if "Average Fried Parameter" in results.colnames:
+            r_none_values = np.array(results["Average Fried Parameter"] == None)
+            if r_none_values.any():
+                results["Average Fried Parameter"][r_none_values] = np.nan
```
Cadair, line 104: Do we also need to change the type of the underlying column in this case? Is it not an object array because of the None's and we want it to be a float?

PR #610, opened 2025-08-18
use new pytest asdf plugin

braingram:
Update test dependencies to use new https://pypi.org/project/pytest-asdf-plugin/ which is a feature-compatible replacement for the bundled asdf pytest plugin (which will soon be deprecated).

For context the pytest-asdf-plugin is responsible for finding and converting schema files into "tests" (to check the schema against the metaschema and validated any contained "examples"). Since both the bundled and now external plugins are feature-compatible there should be no change in test number (or new failures, etc).

On main 443 tests passed:
https://github.com/DKISTDC/dkist/actions/runs/17036230456/job/48289274225#step:10:639

With this PR 443 tests passed:
https://github.com/DKISTDC/dkist/actions/runs/17048153911/job/48329094771?pr=610#step:10:638

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ dynamic = ["version"]
 [project.optional-dependencies]
 tests = [
   "pytest",
+  "pytest-asdf-plugin",
   "pytest-doctestplus",
   "pytest-cov",
   "pytest-remotedata",


braingram:
> Thanks!
> 
> Out of curiosity, does the new package not conflict with the one bundled in asdf?

Not at the moment. Each has checks to only register itself when they won't break the other. So with:
- an old (pre-4.4.0) asdf is installed, pytest-asdf-plugin will not do anything
- if asdf 4.4.0 is installed it will not register the bundled plugin if pytest-asdf-plugin is installed

With asdf 4.4.0 and pytest-asdf-plugin installed the new external plugin will run. In all other cases the bundled plugin will run.

The plan is to:
- deprecate the bundled plugin in asdf 4.5.0
- remove the bundled plugin in asdf 5.0.0
- remove the check for the bundled plugin in pytest-asdf-plugin after a few years (or if they become incompatible)

Cadair:
Thanks!

PR #626, opened 2025-08-26
TiledDataset plot tweaks

SolarDrew:
Optionally hides the tick labels on the interior axes of a Tiled Dataset plot. closing #620 . Also fixes the fact that axis labels weren't being shown. Also also lots of ruff for some reason.


dkist/dataset/tiled_dataset.py
```diff
@@ -333,10 +342,16 @@ def plot(
                 if swap_tile_limits in ["y", "xy"]:
                     ax.invert_yaxis()
 
+                if col == row == 0:
+                    xlabel, ylabel = self._get_axislabels(ax)
                 ax.set_ylabel(" ")
                 ax.set_xlabel(" ")
```
Cadair, line 348: These lines would be better off using the wcsaxes API for consistency. https://docs.astropy.org/en/stable/visualization/wcsaxes/ticks_labels_grid.html#axis-labels

Cadair (approved):

PR #690, opened 2026-02-12
Make slicing Inversions also slice profiles

SolarDrew:


SolarDrew:
I think this is ready for review, although I'm sure I've missed something. I still don't know what's up with the `ResourceWarning` in the figure tests, and I don't know why the docs are whinging about `pkg_resources`.

dkist/dataset/inversion.py
```diff
@@ -157,8 +157,24 @@ def __str__(self):
 
     def __getitem__(self, aslice):
         new_inv = super().__getitem__(aslice)
-        new_inv.profiles = self.profiles
-
+        # If the keys are the same then it was a data slice so we should slice the profiles too
+        if hasattr(new_inv, "keys") and self.keys() == new_inv.keys():
+            # First we need to know which axes are common to the Inversion and the Profiles
+            i_ax = self.meta["axes"]
+            p_ax = self.profiles.meta["axes"]
```
Cadair: I think we should try and use `aligned_axes` here? Would that work. I'd rather not rely on the contents of meta?

Cadair:
rebasing / merging should fix the docs #692

dkist/dataset/tests/test_inversion.py
```diff
@@ -38,9 +39,34 @@ def test_get_item(inversion):
     assert inversion.profiles == sliced_inv.profiles
 
 
+@pytest.mark.parametrize("slice", [np.s_[100:, 10:20], np.s_[10:], np.s_[0], np.s_[:, :10], np.s_[:, :, :10]])
+def test_slice_all(inversion, slice):
```
Cadair: I think we need to add a test of slicing an already sliced inversion.

PR #695, opened 2026-02-23
Fix some bugs

SolarDrew:
- `TiledDataset.flat.tiles_shape` was broken
- requests hates chardet >= 6.0.0

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -57,6 +57,8 @@ filterwarnings =
     ignore:ERFA function "utcut1"*
     ignore:Tried to get polar motions for times after IERS data is valid*
     ignore:leap-second auto-update failed due to the following exception.*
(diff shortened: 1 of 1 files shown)


dkist/dataset/tiled_dataset.py
```diff
@@ -265,6 +265,10 @@ def tiles_shape(self) -> list[tuple[int, ...]]:
         """
         The shape of each individual tile in the TiledDataset.
         """
+        if len(self.shape) == 1:
+            # Need to not nest iteration if we've flattened the TieldDataset
```
Cadair:
```suggestion
            # Need to not nest iteration if we've flattened the TiledDataset
```

Cadair (approved):

PR #713, opened 2026-04-30
Adding functionality to represent non linear spectral dispersion

A-Derks:
**Background:**
DKIST instruments (Cryo SP, DL-NIRSP, ViSP) produce FITS headers with CTYPE values of AWAV-GRA, indicating a non-linear pixel-to-wavelength transform based on the grating/grism equation described in Greisen et al. (2006, FITS WCS Paper 3, sections 5.1.2 and 5.2). The PV{n}_m keywords encode the physical grating parameters.

The gwcs library (via gwcs#745) provides two models for evaluating the FITS grism transform: RefractedAngleSineModel, which computes the pixel-dependent refracted-angle sine from the detector geometry and grating parameters, and WavelengthFromGrismEquation, which converts the incident-angle sine and refracted-angle sine into wavelength using the FITS grism formalism. WavelengthFromGratingEquation remains the classic two-input grating-equation model and is not used by this PR. This PR provides the builder function that wires the FITS grism models together using the parameters extracted from the FITS header.

**What changed:**

- dkist/wcs/models.py

    - build_grating_spectral_transform — the core implementation. Given the full set of grating/grism parameters, it:

        1. Computes the constant sin(alpha_in) directly from incident_angle and wraps it in a Const1D
        2. Constructs a RefractedAngleSineModel (from gwcs.spectroscopy) with the detector geometry and grating parameters baked in
        3. Constructs a WavelengthFromGrismEquation (from gwcs.spectroscopy) with the grism parameters
        4. Assembles the compound model: Mapping((0, 0)) | (Const1D(alpha_in) & RefractedAngleSineModel(...)) | WavelengthFromGratingEquation(...)

All four optional grism parameters (refractive_index, refractive_index_derivative, out_of_plane_angle, camera_angle) default to their identity values, so the transform degrades to a standard grating equation when they are absent from the header.

RefractedAngleSineModel and WavelengthFromGrismEquation are defined in gwcs.spectroscopy (added via gwcs#745), which also provides the ASDF converter that serializes them. This PR imports both directly from gwcs rather than defining or serializing a local model in dkist.

**Notes:**

- The CI test failures are due to the gwcs dependency: RefractedAngleSineModel and WavelengthFromGratingEquation do not exist in any released gwcs version yet. Tests will pass once gwcs#745 merges and the dependency is pinned or updated.

- The CodSpeed regression (-6.9% on test_plot_dataset) is unrelated to the spectral transform code — it affects a plotting benchmark and is likely noise or a pre-existing issue.


dkist/wcs/models.py
```diff
@@ -31,10 +32,115 @@
     "VaryingCelestialTransform2D",
     "VaryingCelestialTransform3D",
     "generate_celestial_transform",
+    "generate_grating_spectral_transform",
     "varying_celestial_transform_from_tables",
 ]
 
 
+def _grating_equation_constants(
```
eigenbrot: Type-hints, please

dkist/wcs/models.py
```diff
@@ -31,10 +32,115 @@
...
+    refracted_angle_sine = _refracted_angle_sine_model(
+        reference_pixel=reference_pixel,
+        dispersion=dispersion,
+        reference_refracted_angle=reference_refracted_angle,
+        theta=theta,
+        angle_slope=angle_slope,
+    )
+    alpha_in = m.Const1D(amplitude=adjusted_incident_angle_sine)
+    wavelength_from_grating = WavelengthFromGratingEquation(
+        groove_density=wavelength_parameter / spectral_order,
```
eigenbrot: If  I'm reading the math right, I think `groove_density = wavelength_parameter / spectral_order` only in the case where `refractive_index_derivative = 0` and `out_of_plan_angle = 0`. Let's just set `groove_density` to the input `grating_density`, which is the same thing.

dkist/wcs/models.py
```diff
@@ -31,10 +32,115 @@
...
+):
+    alpha = incident_angle.to(u.rad)
+    epsilon = out_of_plane_angle.to(u.rad)
+    theta = camera_angle.to(u.rad)
+
+    grism_parameter = (grating_density * spectral_order) / np.cos(epsilon)
+    adjusted_incident_angle_sine = (
+        refractive_index - (refractive_index_derivative * reference_wavelength).decompose()
+    ) * np.sin(alpha)
+    wavelength_parameter = grism_parameter - refractive_index_derivative * np.sin(alpha)
```
eigenbrot: Why is this variable called `wavelength_parameter`? When reading through these functions this name caused me a lot of confusion because I thought it had something to do with `reference_wavelength` or just wavelength in general. The units are 1 / m, which doesn't feel like a wavelength to me.

Based on my comment down in `generate_grating_spectral_transform` I'd argue we don't actually need this variable outside of this function. Given that, we could sidestep any naming issue by just removing this variable altogether and writing out `angle_slope` with the other variables directly.

dkist/wcs/models.py
```diff
@@ -31,10 +32,115 @@
...
+    )
+    angle_slope = wavelength_parameter / (
+        np.cos(reference_refracted_angle) * np.cos(theta) ** 2
+    )
+
+    return (
+        adjusted_incident_angle_sine,
+        wavelength_parameter,
+        reference_refracted_angle,
+        theta,
```
eigenbrot: I don't think we need to return `theta` from this function. It's just a variable rename of the input parameter `camera_angle`, which (I think) just muddies the path of variables in `generate_grating_spectral_transform`. I.e., it makes it seem like `theta` is meaningfully different from `camera_angle`, which it isn't.

True, a `.to(u.rad)` has happened, but the return value is still a `Quantity` and so, e.g., `np.sin` will work correctly in any case.

dkist/wcs/models.py
```diff
@@ -31,10 +32,115 @@
...
+    ) = _grating_equation_constants(
+        reference_wavelength=reference_wavelength,
+        grating_density=grating_density,
+        spectral_order=spectral_order,
+        incident_angle=incident_angle,
+        refractive_index=refractive_index,
+        refractive_index_derivative=refractive_index_derivative,
+        out_of_plane_angle=out_of_plane_angle,
+        camera_angle=camera_angle,
+    )
```
eigenbrot: It appears this function (`_grating_equation_constants`) exists to save a little bit of code repetition when computing some of the inputs to `_refracted_angle_sine_model`, `alpha_in`, and `WavelengthFromGratingEquation`, but, in doing so, it separates the inputs to `generate_grating_spectral_transform` from their actual usage and adds another layer of variable names and calculations. I think this added layer causes confusion (see my comments above) and obfuscates the flow of information through `generate_grating_spectral_transform`.

My suggestion would be to remove `_grating_equation_constants` entirely and rewrite `_refracted_angle_sine_model` to accept `grating_density`, `spectral_order`, `refractive_index`, and `reference_wavelength`. `adjusted_incident_angle_sine` could be computed directly in the function body.

dkist/wcs/models.py
```diff
@@ -31,10 +32,115 @@
...
+    return refracted_angle_sine()
+
+
+def generate_grating_spectral_transform(
+        reference_pixel,
+        reference_wavelength,
+        dispersion,
+        grating_density,
+        spectral_order,
+        incident_angle,
```
eigenbrot: type-hints, please

dkist/wcs/models.py
```diff
@@ -31,10 +32,115 @@
...
+        theta,
+        angle_slope,
+    )
+
+
+def _refracted_angle_sine_model(
+        reference_pixel, dispersion, reference_refracted_angle, theta, angle_slope
+):
+    @custom_model
+    def refracted_angle_sine(pixel):
```
eigenbrot: We need a docstring either on `_refacted_angle_sine_model` or `refracted_angle_sine`.

dkist/wcs/models.py
```diff
@@ -31,10 +32,115 @@
...
+        angle_slope=angle_slope,
+    )
+    alpha_in = m.Const1D(amplitude=adjusted_incident_angle_sine)
+    wavelength_from_grating = WavelengthFromGratingEquation(
+        groove_density=wavelength_parameter / spectral_order,
+        spectral_order=spectral_order,
+        name="Spectral",
+    )
+
+    return m.Mapping((0, 0)) | (alpha_in & refracted_angle_sine) | wavelength_from_grating
```
eigenbrot: It looks like the `model` produced by `WavelengthFromGratingEquation` [does expect](https://gwcs.readthedocs.io/en/latest/api/gwcs.spectroscopy.WavelengthFromGratingEquation.html) the inputs to be the sine of the input and output angles. This is what is correctly written here, but the variable names make it look like we're passing a raw angle (`alpha_in`) and the sine of an angle (`refracted_angle_sine`). However we choose to name these variables we should use the same convention for both.

We might also consider referring to these angles as "input" and "output" instead of "alpha/incident" and "refracted" to be clear about the connection to the language used in `WavelengthFrameGratingEquation`, but I'm also fine sticking with "incident" and "refracted".

dkist/wcs/models.py
```diff
@@ -31,10 +32,115 @@
...
+    epsilon = out_of_plane_angle.to(u.rad)
+    theta = camera_angle.to(u.rad)
+
+    grism_parameter = (grating_density * spectral_order) / np.cos(epsilon)
+    adjusted_incident_angle_sine = (
+        refractive_index - (refractive_index_derivative * reference_wavelength).decompose()
+    ) * np.sin(alpha)
+    wavelength_parameter = grism_parameter - refractive_index_derivative * np.sin(alpha)
+    reference_refracted_angle = np.arcsin(
+        (grism_parameter * reference_wavelength).decompose() - refractive_index * np.sin(alpha)
```
eigenbrot: Is the `.decompose()` call necessary here? It looks like the results of this calculation stay in `Quantity` land for the duration of their life.

dkist/wcs/models.py
```diff
@@ -31,10 +32,115 @@
...
+        refractive_index,
+        refractive_index_derivative,
+        out_of_plane_angle,
+        camera_angle,
+):
+    alpha = incident_angle.to(u.rad)
+    epsilon = out_of_plane_angle.to(u.rad)
+    theta = camera_angle.to(u.rad)
+
+    grism_parameter = (grating_density * spectral_order) / np.cos(epsilon)
```
eigenbrot: Not to be That Guy, but Greisen already defines something called the "grism parameter" in Eq. 75, which is different than the definition here.

dkist/wcs/models.py
```diff
@@ -31,10 +32,115 @@
...
+        out_of_plane_angle=0 * u.deg,
+        camera_angle=0 * u.deg,
+) -> CompoundModel:
+    """
+    Generate a one-dimensional FITS ``-GRA``/``-GRI`` spectral transform.
+
+    This function handles the FITS grating/grism intermediate-angle terms and
+    delegates the final wavelength calculation to
+    ``gwcs.spectroscopy.WavelengthFromGratingEquation``.
+    """
```
eigenbrot, line 63: It would be helpful to state that the input and output angles are computed based on the Greisen paper and then provide a reference here in the docstring. I would recommend either linking the [html version](https://www.aanda.org/articles/aa/full/2006/05/aa3818-05/aa3818-05.html) or the [ADS link](https://ui.adsabs.harvard.edu/abs/2006A%26A...446..747G/abstract) instead of the pdf hosted on Greisen's personal website.

If we want to go all-in on the paper connection we could match all of our variables names to names used in the paper, but that's probably overkill.

dkist/wcs/tests/test_models.py
```diff
@@ -52,6 +54,36 @@ def test_generate_celestial_unitless():
...
+    header = {
+        "CTYPE1": "AWAV-GRA",
+        "CUNIT1": "nm",
+        "CRPIX1": 218,
+        "CRVAL1": 854.1738582455826,
+        "CDELT1": 0.0022975580183395555,
+        "PV1_0": 23000.0,
+        "PV1_1": 90,
+        "PV1_2": 65.696,
+    }
```
eigenbrot, line 72: Looking at the other tests in this module, I think it's fine to test only a single case here, but we should at least provide non-default values for all of the parameters. Specifically `PV1_3/refractive_index`, `PV1_4/refractive_index_derivative`, `PV1_5/out_of_plan_angle`, and `PV1_6/camera_angle`. This would've caught the assumptions inherent in `groove_density=wavelength_parameter / spectral_order`.

dkist/wcs/tests/test_models.py
```diff
@@ -52,6 +54,36 @@ def test_generate_celestial_unitless():
...
+        reference_pixel=header["CRPIX1"] - 1,
+        reference_wavelength=header["CRVAL1"] * u.nm,
+        dispersion=header["CDELT1"] * u.nm / u.pix,
+        grating_density=header["PV1_0"] / u.m,
+        spectral_order=header["PV1_1"] * u.one,
+        incident_angle=header["PV1_2"] * u.deg,
+    )
+
+    pixels = np.array([0, 100, 217, 300, 511], dtype=float)
+    expected = WCS(header).all_pix2world(pixels, 0)[0] * u.m
```
eigenbrot: As mentioned on the related `dkist-inventory` PR, a disadvantage of `all_pix2world` is that the results are unit-less and require us to apply the assumed-correct unit to its output. (It also has this weird signature and output that require non-obvious 0 and [0]).

Instead, I’d suggest using `WCS(header).spectral.pixel_to_world(pixels)`, which will have proper units based on the CUNIT key. Then `expected.to_value(u.nm)` will work directly without needing to assume anything in the meantime.

eigenbrot (changes requested):

dkist/wcs/models.py
```diff
@@ -31,10 +32,115 @@
...
+        reference_refracted_angle,
+        theta,
+        angle_slope,
+    )
+
+
+def _refracted_angle_sine_model(
+        reference_pixel, dispersion, reference_refracted_angle, theta, angle_slope
+):
+    @custom_model
```
Cadair: Can you refactor this into class form, I think that with it like this it will be hard to add the required asdf schemas.

dkist/wcs/models.py
```diff
@@ -31,10 +32,115 @@
...
+            np.arctan(-np.tan(theta) + wavelength_offset * angle_slope)
+            + reference_refracted_angle
+            + theta
+        )
+        return np.sin(output_angle)
+
+    return refracted_angle_sine()
+
+
+def generate_grating_spectral_transform(
```
Cadair: I think this function can (and probably should?) live in dkist-inventory. The model needs to be in here so that it can be imported at ASDF read time.

dkist/wcs/tests/test_models.py
```diff
@@ -52,6 +54,36 @@ def test_generate_celestial_unitless():
     assert u.allclose(shift1.offset, 0)
 
 
+def test_generate_grating_spectral_transform():
```
Cadair: Can you also add tests for just the new model in isolation? I also suspect this test should probably be in dkist-inventory and not here.

Cadair (commented):
Sorry to keep bouncing parts of this code between here and dkist-inventory!

The thing this PR is missing is the [asdf converter](https://github.com/DKISTDC/dkist/blob/main/dkist/io/asdf/converters/models.py), [schema](https://github.com/DKISTDC/dkist/blob/main/dkist/io/asdf/resources/schemas/asymmetric_mapping_model-1.2.0.yaml) and corresponding entries in the manifest, tests etc.

This is *a thing*, so this should be the last thing to do, but it will be needed to successfully test a roundtrip to ASDF file for this (which there should be a test of in dkist-inventory). We do have a helper script for this process which is [here](https://github.com/DKISTDC/dkist/blob/main/tools/update_schema.py).

You should (for this PR) be able to run it like this:
```shell
$ pipx run tools/update_schema.py --manifest dkist-wcs refracted-angle-sine-model
```
~However, it seems like this is bugged, so we'll have to fix it first.~ It's now fixed enough it runs, it should get you started.

dkist/wcs/models.py
krislarsnso: "grism constant" would be an OK alternative

dkist/wcs/models.py
krislarsnso: I agree with Arthur, but if you do want to name these parts of the grism equation (Greisen eq 70), then I would stick with names that imply units.  You already do that with `adjusted_incident_angle_sine`, so you could name this part with something that includes `...inverse_wavelength`.

dkist/wcs/models.py
```diff
@@ -31,10 +32,115 @@
...
+
+    grism_parameter = (grating_density * spectral_order) / np.cos(epsilon)
+    adjusted_incident_angle_sine = (
+        refractive_index - (refractive_index_derivative * reference_wavelength).decompose()
+    ) * np.sin(alpha)
+    wavelength_parameter = grism_parameter - refractive_index_derivative * np.sin(alpha)
+    reference_refracted_angle = np.arcsin(
+        (grism_parameter * reference_wavelength).decompose() - refractive_index * np.sin(alpha)
+    )
+    angle_slope = wavelength_parameter / (
```
krislarsnso: I would like to rename this.  If we use the name "grism parameter" for Γ (i.e. rename Gm/cosϵ) then this is the slope of the grism parameter wrt wavelength and wrt the distance along the detector (ξ in the paper) scaled.  Γ itself is dimensionless.  I think `grism_parameter_slope` would be better.  Or maybe `grism_parameter_per_wavelength`, which would make its use in `_refracted_angle_sine_model` more clear.

dkist/wcs/models.py
```diff
@@ -31,10 +32,115 @@
...
+        spectral_order,
+        incident_angle,
+        refractive_index,
+        refractive_index_derivative,
+        out_of_plane_angle,
+        camera_angle,
+):
+    alpha = incident_angle.to(u.rad)
+    epsilon = out_of_plane_angle.to(u.rad)
+    theta = camera_angle.to(u.rad)
```
krislarsnso: If these are already Quantities, then I don't think we need to convert them to radians.  In this function, we only ever do trig on them, not use them directly in algebra, so the conversion isn't necessary.

dkist/wcs/models.py
krislarsnso: @eigenbrot, see my [comment down on line 84](https://github.com/DKISTDC/dkist/pull/713/changes#r3190311133).

dkist/wcs/models.py
```diff
@@ -31,10 +32,115 @@
...
+def _refracted_angle_sine_model(
+        reference_pixel, dispersion, reference_refracted_angle, theta, angle_slope
+):
+    @custom_model
+    def refracted_angle_sine(pixel):
+        wavelength_offset = ((pixel - reference_pixel) * u.pix) * dispersion
+        output_angle = (
+            np.arctan(-np.tan(theta) + wavelength_offset * angle_slope)
+            + reference_refracted_angle
+            + theta
```
krislarsnso: Here is where `theta` aka `camera_angle` _does_ need to be in radians.  If it isn't passed through, the conversion should happen here.

dkist/wcs/models.py
```diff
@@ -31,10 +32,115 @@
...
+
+
+def _refracted_angle_sine_model(
+        reference_pixel, dispersion, reference_refracted_angle, theta, angle_slope
+):
+    @custom_model
+    def refracted_angle_sine(pixel):
+        wavelength_offset = ((pixel - reference_pixel) * u.pix) * dispersion
+        output_angle = (
+            np.arctan(-np.tan(theta) + wavelength_offset * angle_slope)
```
krislarsnso: Renaming `angle_slope` to something like `grism_parameter_per_wavelength` (see [comment here](https://github.com/DKISTDC/dkist/pull/713/changes#r3190154624)) then makes this line clearer about what it is doing.

dkist/wcs/models.py
krislarsnso: It looks like the [transition from ADS to SciX](https://scixplorer.org/adstoscix/) will happen soon, so we could use the [SciX link](https://scixplorer.org/abs/2006A%26A...446..747G/abstract).

IMO matching _more_ variable names would not be overkill.

dkist/wcs/models.py
krislarsnso: +1 to same name convention for both

krislarsnso (changes requested):

dkist/wcs/models.py
krislarsnso: But here's the part I don't get.  If groove density is just G, then where is the denominator in Greisen eq 70?  I'm looking at step 3 on page 756: "Compute the wavelength \lambda from \Gamma via Eq. 70"  That equation has the two numerator parts, an adjusted sin \alpha and a sin \gamma, which we pass in the return of this function.  The denominator in eq 70 is what is called `wavelength_parameter` above.  Are we supposed to scale the sin(angles) before passing them to the model?

dkist/wcs/models.py
```diff
@@ -35,6 +35,78 @@
 ]
 
 
+class _refracted_angle_sine_model:
```
Cadair: This should be refactored to be a subclass of Model like some of the other classes in this file, and not use `@custom_model`, like this example: https://docs.astropy.org/en/stable/modeling/new-model.html#a-full-example-of-a-linemodel

dkist/wcs/models.py
```diff
@@ -35,6 +35,78 @@
...
+        camera_angle: u.Quantity = 0 * u.deg,
+) -> CompoundModel:
+        """
+        Build a one-dimensional FITS ``-GRA``/``-GRI`` spectral transform.
+
+        This wrapper preserves the historical public API in ``dkist`` while
+        delegating the implementation to ``dkist-inventory`` when it is
+        available.
+        """
+
```
eigenbrot, line 90: It looks like we're missing the actual code here?

dkist/wcs/tests/test_models.py
A-Derks: Added in there

krislarsnso (approved):

changelog/713.feature.rst
```diff
@@ -0,0 +1 @@
+Added `dkist.wcs.models.build_grating_spectral_transform()` and updated `dkist.wcs.models.generate_grating_spectral_transform()` to build FITS ``-GRA``/``-GRI`` spectral transforms by composing a constant incident-angle term, `~dkist.wcs.models.refracted_angle_sine_model`, and `~gwcs.spectroscopy.WavelengthFromGratingEquation` using the FITS-facing grating parameters.
```
eigenbrot: It looks like this needs to be updated to reflect the new location of the refracted angle model.

dkist/wcs/models.py
```diff
@@ -30,11 +31,115 @@
...
     "VaryingCelestialTransform3D",
+    "build_grating_spectral_transform",
     "generate_celestial_transform",
+    "generate_grating_spectral_transform",
+    "refracted_angle_sine_model",
     "varying_celestial_transform_from_tables",
 ]
 
 
+class refracted_angle_sine_model(RefractedAngleSineModel):
```
eigenbrot: Can't we just use `gwcs.spectroscopy.RefractedAngleSineModel` directly? It looks like this class is no longer needed.
I understand from the docstring that there is a need to keep stable paths, but wasn't  `refracted_angle_sine_model` introduced in this PR? There should be no datasets that reference it yet, right?

dkist/wcs/models.py
```diff
@@ -30,11 +31,115 @@
...
+    reference_pixel: float,
+    reference_wavelength: u.Quantity,
+    dispersion: u.Quantity,
+    groove_density: u.Quantity,
+    spectral_order: u.Quantity,
+    incident_angle: u.Quantity,
+    refractive_index: u.Quantity = 1 * u.one,
+    refractive_index_derivative: u.Quantity = 0 / u.m,
+    out_of_plane_angle: u.Quantity = 0 * u.deg,
+    camera_angle: u.Quantity = 0 * u.deg,
```
eigenbrot, line 50: Just out of curiosity, how did we decide to draw the line at which inputs have default values and which don't?

dkist/wcs/models.py
```diff
@@ -30,11 +31,115 @@
...
+        refractive_index=refractive_index,
+        refractive_index_derivative=refractive_index_derivative,
+        out_of_plane_angle=out_of_plane_angle,
+        camera_angle=camera_angle,
+    )
+
+    return m.Mapping((0, 0)) | (alpha_in & alpha_out) | model
+
+
+def generate_grating_spectral_transform(
```
eigenbrot: What is the important difference between `generate_grating_spectral_transform` and `build_grating_spectral_transform`? 

This function here literally just passes all of its inputs directly to `build_grating_spectral_transform` and I don't see why it's necessary. Can't we just call `build_grating_spectral_transform` directly?

eigenbrot (changes requested):

changelog/713.feature.rst
A-Derks: Good catch. Updated.

dkist/wcs/models.py
A-Derks: Yeah, there really wasn’t an important difference between generate_grating_spectral_transform() and build_grating_spectral_transform(). The former was just a pass-through wrapper, so I removed it and kept build_grating_spectral_transform() as the public entry point.

dkist/wcs/models.py
A-Derks: Required FITS definition inputs are defined and optional grism correction terms aren't. I also added more documentation about this choice.

dkist/wcs/models.py
A-Derks: Yes you are right. I have removed this old unused class and now use gwcs.spectroscopy.RefractedAngleSineModel directly.

dkist/wcs/models.py
eigenbrot: I think there's some confusion here. To lay it all out, for the `dkist` package we can think about three separate types of wavecal parameters (`variable_name` and (FITS_KEY)):

Parameters "required" by FITS
--------------------------------------
(really only "required" in the sense that these are the minimum needed to generate _any_ kind of wavelength axis; i.e., a FITS file missing these keys is still totally valid)

- `reference_pixel` (CRPIX*k*)
- `reference_wavelength` (CRVAL*k*)
- `dispersion` (CDELT*k*)

Non-linear Wavelength Solution Parameters a la Greisen
------------------------------------------------------------------------
In addition to the linear parameters above, the full non-linear model supports these parameters

- `groove_density` (PV*k*_0)
- `spectral_order` (PV*k*_1)
- `incident_angle` (PV*k*_2)
- `refractive_index` (PV*k*_3)
- `refractive_index_derivative`(PV*k*_4)
- `out_of_plane_angle` (PV*k*_5)
- `camera_angle` (PV*k*_6)

Parameters currently used/fit in `solar-wavelength-calibration`
-----------------------------------------------------------------------------
The subset of Greisen's full parameter set that is currently used by `solar-wavelength-calibration`. As a result all DKIST spectroscopic data will have these parameters in their headers

- `groove_density` (PV*k*_0)
- `spectral_order` (PV*k*_1)
- `incident_angle` (PV*k*_2)

The point is that _all_ of these parameters live in FITS headers if they are used, and that only the first three are "required" in any sense.

Having said all that, it sounds like the parameters currently supported by `solar-wavelength-calibration` are required arguments to this function and the remaining four Greisen parameters are kwargs with sensible defaults. This feels pretty reasonable to me, but we need to be careful with the language. FWIW, I don't think the addition to the docstring is necessary. I was just curious about choice of arg/kwarg and I think I understand that better now.

A-Derks: Removed the addition to the docstring.

eigenbrot (approved):
Approving, but the current PR description is out of date

fraserwatson (approved):

PR #721, opened 2026-05-07
Updated quality report call to download json and pdf

deege:
Issue: https://nso.atlassian.net/browse/DCS-4635

Adds format support to DKISTFileManager.quality_report() so callers can download either the existing PDF report or the new JSON report.

Preserves default PDF behavior for quality_report()
Supports format="pdf" and format="json"
Sends Accept: application/json for JSON requests, matching metadata-object-streamer negotiation
Raises ValueError for unsupported formats
Adds focused tests for default/PDF, JSON, and invalid format behavior


dkist/io/file_manager.py
```diff
@@ -162,9 +170,20 @@ def quality_report(self, path: str | os.PathLike | None = None, overwrite: bool
...
         url = f"{self._metadata_streamer_url}/quality?datasetId={self._dataset_id}"
         if path is None and self.basepath:
             path = self.basepath
+        if normalized_format == "json":
+            dl = Downloader()
+            dl.enqueue_file(
+                url, path=path, overwrite=overwrite, headers={"Accept": "application/json"}
+            )
+            return dl.download()
         return Downloader.simple_download([url], path=path, overwrite=overwrite)
```
Cadair: can we remove this and have one download path where we conditionally add the header instead?

Cadair (approved):

PR #735, opened 2026-07-03
Give each array built by stack_loader_array a unique dask graph layer name

astrofrog:
Every array built by ``stack_loader_array`` uses the hard-coded dask graph layer name load_files, so any two ``FileManager``-backed arrays have identical dask graph keys. Dask identifies arrays by these keys, so combining two such arrays into a single graph silently deduplicates their file-loading tasks and uses one array's data for all of them:

```python
In [1]: import dkist

In [2]: from dask import array as da

In [3]: import numpy as np

In [4]: ds = dkist.load_dataset("EXVXQL")

In [5]: tiles = list(ds.flat)

In [6]: stacked = da.stack([tiles[0].data[:10,:10,:10], tiles[1].data[:10,:10,:10]]).compute()

In [7]: np.allclose(stacked[0], stacked[1], equal_nan=True)
Out[7]: True  # should be False!
```

Computing each array separately gives correct results, so this only shows up when arrays are combined lazily - stacking tiles of a TiledDataset, arithmetic between datasets, or deferred mosaicking (this was found via ``reproject_and_coadd`` in the reproject package with a new dask mode, which produced a mosaic with one tile's data repeated in every tile position).

The fix gives each array a unique layer name, matching how dask names its own collections. A regression test stacks two single-file slices and checks each keeps its own data.


SolarDrew (approved):
Looks good, thanks for finding this.

PR #747, opened 2026-08-06
Rename the `figure=` kwarg of `TiledDataset.plot` to `fig=`

### LINKED ISSUE #593: `NDCube.plot` (for >2D) accepts `fig=` and `TiledDataset.plot` accepts `figure=`
☹️ 🐼 

Bring TiledDataset in line with NDCube and add a decent deprecation period to it.

sr-dash:
Fixes #593.

`TiledDataset.plot()` took the figure to draw into as `figure=`, while `NDCube.plot()` (and therefore `Dataset.plot()` for >2D data) takes it as `fig=`. This renames the `TiledDataset.plot()` keyword to `fig` and deprecates `figure` with `astropy.utils.decorators.deprecated_renamed_argument`, so existing code carries on working (with a `DKISTDeprecationWarning`) until the deprecation period is up.

The parameter keeps its position in the signature, so anyone passing it positionally is unaffected.

- Renamed the keyword and updated the docstring, the tutorial (`docs/tutorial/7_visualization.md`) and the VBI extents example.
- Added two tests: one that `fig=` returns the figure it was given, and one that `figure=` still works but warns.

I set the deprecation version to `1.19` on the assumption that's the next release — happy to change it.

Two things I wasn't sure about and left alone, let me know if you'd like them in this PR:

- `Inversion.plot()` and `Profiles.plot()` also take `figure=`. They're experimental, but renaming them too would keep the whole package consistent.
- ~~I haven't added a `changelog/` entry yet~~ — added as `747.removal.rst`.


Cadair:
Hi! Thanks for the PR, I think it'd be good to change `Inversion` as well.

SolarDrew (commented):
I agree, I'd like to have consistency with the `Inversion` and `Profiles` classes as well. Otherwise this looks good, thanks for fixing it.

PR #748, opened 2026-08-06
Add `qualityDataObjectKey` to the inventory pretty names

### LINKED ISSUE #649: Add `qualityDataObjectKey` to inventory pretty names
New key has been added to inventory, so we should humanize it.

sr-dash:
Fixes #649.

Adds `"qualityDataObjectKey": "Quality Data Filename"` to `INVENTORY_KEY_MAP`, so the key is translated in search results and gets a `quality_data_filename` path interpolation key, matching the existing `qualityReportObjectKey` -> `"Quality Report Filename"`.

I checked what the API actually returns for the key before picking the name:

```
'qualityReportObjectKey' = 'pid_1_123/BDNEK/BDNEK_quality_report.pdf'
'qualityDataObjectKey'   = 'pid_1_123/BDNEK/BDNEK_quality_data.json'
```

so "Quality Data Filename" seemed like the natural counterpart, but I'm happy to rename it if you'd prefer something that makes the PDF/JSON distinction more explicit.

Extended `test_humanize_loop` to round-trip the new key rather than adding a separate test, since it's exactly what that test is for. The fixed keymap in `test_path_format_table` is deliberately untouched — it's independent of `INVENTORY_KEY_MAP` so adding entries doesn't churn it.

Two things I noticed but left out of scope, tell me if you'd like either as a follow-up:

- `tileCount` is also returned by the search API and isn't in `INVENTORY_KEY_MAP`.
- `DKISTFileManager._download_dataset` adds `qualityReportObjectKey` to the Globus transfer file list; it doesn't add the new quality data file. That's a behaviour change rather than a naming one, so I've not touched it.

### DIFF
--- a/changelog/748.feature.rst
+++ b/changelog/748.feature.rst
@@ -1,0 +1,1 @@
+Add a human readable name for the ``qualityDataObjectKey`` inventory key, which is displayed as "Quality Data Filename" and can be used as the ``{quality_data_filename}`` path interpolation key.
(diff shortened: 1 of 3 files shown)


Cadair (approved):
Thanks!

PR #751, opened 2026-08-13
Fix ravel model behaviour to properly clip OOB coordinates

SolarDrew:
astropy v8.0.1 broke our coordinate transforms because the Ravel model wasn't clipping the lower bounds of the coordinates so it was passing numpy negative numbers. This finds the locations of the coords array where the coords are invalid and masks them out of the ravelled coords array.


Cadair (approved):
The tests are perfect right? *right?*

Cadair (changes requested):
oh wait, this bug needs regression tests.

Cadair (approved):

